Issue #44, opened 2020-03-20
Multiple design alternatives for parents


jameslibscie:
**User story**
As a Hypergraph user
I want an intuitive way of adding a next step (child module) to my work
and viewing and navigating between previous steps (parents) of my work
so that I add my content to Hypergraph correctly
and can get an overview of the connections between my different pieces of work

**Acceptance criteria**
- [ ] Multiple designs for / implementations of the `parents` concept that:
  - [x] visually strengthen the parent / child or step-by-step relationship between different modules
  - [x] offer a way to create a next step / add child content for an existing content module (https://github.com/hypergraph-xyz/desktop/pull/46)

__edit__ by @juliangruber: add hierarchy to acceptance criteria, add (merged) PR

jameslibscie:
Just checked out the version in [commit d7e9420](https://github.com/hypergraph-xyz/desktop/commit/d7e9420b1ae0b9d25a1b9c6cc519728df4cd22cd). I like the design aesthetically, but have some concerns about whether the functionality is self-evident. I could give all kinds of suggestions, but we should really just test it :) Nice to start minimal.

PR #46, opened 2020-03-23
add create content with parent module. #44

juliangruber:
This adds the first part of parent modules: creating a module with a parent, and displaying that parent in the content details screens.

I'm going to base all further parent module pull requests on this one, as they both rely on above mentioned changes.

## Screens

1. Add + (only visible when mouse over)

Screenshot 2020-03-23 at 17 14 09

2. Create content with parent

Screenshot 2020-03-23 at 17 10 38

3. View content with parent

Screenshot 2020-03-23 at 17 10 44


psobrakseaton:
So far I think this is a great solution to the parent module/following-from issue. Can the "open folder" and "export .zip" buttons be moved into the header bar? 

Second, if there is "no parent" can that part of the form disappear completely?

juliangruber:
> Can the "open folder" and "export .zip" buttons be moved into the header bar?

This is independent of this issue, please constrict your review to the changes here. Can you send designs for this, or explain in words how this should look?

> Second, if there is "no parent" can that part of the form disappear completely?

It already only shows when there is a parent:

Screenshot 2020-03-23 at 17 38 54

chartgerink:
For a first implementation w/o designs from @psobrakseaton, it's looking really good :) We'll definitely need to design it out a bit more to clarify some of the concepts.

My only suggestion is to not allow parenting when the content module is unpublished (i.e., no `+` gets displayed then). This is not in the spec per se, but it seems to make sense in my head right now to disallow this from a UX perspective? It's like continuing on unpublished work, it makes it possible that the retracing of steps at a later timepoint gets complicated due to availability issues.

juliangruber:
> My only suggestion is to not allow parenting when the content module is unpublished (i.e., no `+` gets displayed then).

This is unintuitive to me, as you're writing your paper I'd expect to create following content without publishing prior steps first. Please correct me if I'm wrong, but isn't basing drafts on drafts important, especially in uncertain areas?

juliangruber:
> For a first implementation w/o designs from @psobrakseaton, it's looking really good :) We'll definitely need to design it out a bit more to clarify some of the concepts.

Thanks! I didn't want to make any design decisions, so I implemented the simplest possible visual changes.

chartgerink:
I'd need to think about this a bit more in-depth and about the second/third order effects and how they relate to the higher levels of the vision. For now, we can park it (not a huge thing here) and I'll come back to it later through the specs.

psobrakseaton:
I currently don't have an opinion regarding the not showing of a + when there is unpublished content, but I will give it some thought.

As for the implementation of the + icon, I would like to suggest the following, it's purely a visual redesign:

-Upon hover, a 128px bar slides in from the right further reinforcing that you are adding to this module.

-Make that whole area clickable for the sake of accessibility.

 [Here's the Figma board]([](https://www.figma.com/file/5aQDo21KHshfXedOwI3z0o/desktop-follows-from?node-id=61%3A5))

psobrakseaton:
Also, I forgot to say I quickly experimented with moving the bookmark function to the side as well. No need to implement that now.

juliangruber:
> I'd need to think about this a bit more in-depth and about the second/third order effects and how they relate to the higher levels of the vision. For now, we can park it (not a huge thing here) and I'll come back to it later through the specs.

Sounds great! :) If you feel strongly about this, I can also disable the `+` for unpublished content, it's an easy change.

juliangruber:
Merging, as approved by Patch in Slack

jameslibscie:
@juliangruber @chartgerink The CLI currently only allows selecting a published content module afaik, but the specs don't say anything about this. I'll create a specs issue to think about this and then make the products consistent.

PR #50, opened 2020-03-26
add horizontal parent navigation (wip)

juliangruber:


jameslibscie:
I love how easy it is to add new child content this way. I've found a small bug though (which might become irrelevant later according to how we spec the whole `parents` + `publish` thing):
- [ ] When going back and forth between parent/child using the arrows, modules always show as published (green indicator + author name is a link)

Also, I think it's slightly weird that you can have a published module with an unpublished parent. In that case the unpublished parent also shows on the profile if you navigate to the left. This is definitely not something to resolve now, though. Need to consider the combination of those two features very carefully. Also wondering how this will work when you have a parent from a different author. Is it weird to show a parent module from author A on author B's profile? Again, not something we need to resolve now :)

juliangruber:
Closing in favor of https://github.com/hypergraph-xyz/desktop/pull/51

PR #75, opened 2020-05-13
Update issue templates

jameslibscie:


jameslibscie:
Replaced with #76 due to commits being unsigned

PR #78, opened 2020-05-15
Add profile editing

### LINKED ISSUE #70: Multi-line descriptions show as single-line
## Description

If a description contains multiple lines (with newline characters), it gets displayed on a single line.

## To Reproduce

1. Go to "Add content"
2. Add a multi-line description, like:
```
Line 1.
Line 2.
```
3. Save content
4. View content / go to content list

## Expected behavior

The text "Line 1" and "Line 2" are shown on separate lines.

## Actual behavior

The text "Line 1" and "Line 2" is shown on one line.

## Screenshots

image

## Desktop

- OS: Windows
- Version: 0.1.0

juliangruber:
There's quite a bit more in here than just adding profile editing:

- add profile editing
- fix module content margin
- fix input letter spacing
- fix textarea letter spacing and line height
- right align row buttons
- remove row spacers
- fix missing newlines in description. closes #70
- all content blocks have small left margin
- use react element names as css class names for debugging


jameslibscie:
Looks really good!

I have two comments:
- If I enter an invalid title, the indicator just flashes orange. Would it be possible for it to flash red on the appropriate field on error? 300 chars is ridiculously long and empty or whitespace-only titles are also pretty obviously invalid, so I think we can get away with not displaying a clear indication of what the error is
- Right now, the user is only prompted to add a description once Edit profile has been clicked. That leaves no indication as to what the description area is for until they've filled one in. Could we add an "Add a description..." placeholder text that triggers the edit profile action when clicked?

Let's discuss both on Slack, because I'd like Patch's input too

juliangruber:
> * If I enter an invalid title, the indicator just flashes orange. Would it be possible for it to flash red on the appropriate field on error? 300 chars is ridiculously long and empty or whitespace-only titles are also pretty obviously invalid, so I think we can get away with not displaying a clear indication of what the error is

Good catch, I haven't tested validation failures! I implemented this just like you suggested, and I think that works well!

> * Right now, the user is only prompted to add a description once Edit profile has been clicked. That leaves no indication as to what the description area is for until they've filled one in. Could we add an "Add a description..." placeholder text that triggers the edit profile action when clicked?

Great idea! Implemented as requested.

> Let's discuss both on Slack, because I'd like Patch's input too

I just saw this when I finished addressing the points :|

psobrakseaton:
- When entering an invalid title, we should stay in edit mode if they try to save. If they cancel it should revert to the last saved version.
- When editing the text of both the title and the description should have drag to select enabled.
- I would like to be able to reposition the window while dragging in the blank space above the header bar.

juliangruber:
@psobrakseaton Implemented both, however I removed the pulsating of the indicator which made the implementation too complicated

juliangruber:
> * save button text turns from green to gray (instant/dissolve - same transformation as the typing indicator)

A transition here isn't possible as that would then also apply to other state changes, like when you hover a non-disabled button. Animations like that theoretically are implementable, but at a high price. Think of css transitions as globally defined per element, eg `button { transition: color 1s; }`

juliangruber:
> * 32px*32px warning icon appears 16px from the left save button (the icon is export ready)

Shall we shrink the title input then, so that there always is space for the icon? I think we could also do without the icon for now tbh

psobrakseaton:
> > * 32px*32px warning icon appears 16px from the left save button (the icon is export ready)
> 
> Shall we shrink the title input then, so that there always is space for the icon? I think we could also do without the icon for now tbh

Yeah, it may be overkill for right now. I want to approach tooltips next milestone so we have simplified designs, but extra explanations where needed. I think this will be very important with more robust network features.

psobrakseaton:
> > * save button text turns from green to gray (instant/dissolve - same transformation as the typing indicator)
> 
> A transition here isn't possible as that would then also apply to other state changes, like when you hover a non-disabled button. Animations like that theoretically are implementable, but at a high price. Think of css transitions as globally defined per element, eg `button { transition: color 1s; }`

Ah ok, that makes sense. Thanks for the clarification. Lets do without the transition then for the sake of efficiency. I'll defer to your discretion for the state change of the button. In terms of the visual design itself are there any concerns?

psobrakseaton:
Posted a comment in slack, when the line of the description is filled there is no wrapping of the text, it just continues to the right.

I will begin designing a "read more" action, but it doesn't have to be implemented yet.

juliangruber:
> Posted a comment in slack, when the line of the description is filled there is no wrapping of the text, it just continues to the right.

Good catch! I had to undo the auto resizing of the textarea as this is just very difficult to get right. It's always 3 lines high now.

psobrakseaton:
> > Posted a comment in slack, when the line of the description is filled there is no wrapping of the text, it just continues to the right.
> 
> Good catch! I had to undo the auto resizing of the textarea as this is just very difficult to get right. It's always 3 lines high now.

Great! Excited to test the newest version. Would you prefer I tackle the "read more" bit now, or is that a task you'd like to tackle later?

juliangruber:
I don't think this is pressing, users have a simple fix by just shortening their description

psobrakseaton:
Ok, so it now wraps while I am editing, but if I save then it goes back to one line.

jameslibscie:
@psobrakseaton I would have liked to have reviewed this before merging.

juliangruber:
> @psobrakseaton I would have liked to have reviewed this before merging.

I added this for discussion to the milestone doc, I think given the upcoming "deadline" I fell into the old pattern of having to merge before the sprint ends.

PR #85, opened 2020-05-26
Fix open dialog default path. Closes #77

### LINKED ISSUE #77: Open first "upload files" modal to Documents folder
**User story**
As a content creator
I want Hypergraph to open my default documents folder when adding files for the first time
so that I can find my files more quickly

**Acceptance criteria**
- [ ] When opening the upload files modal for the first time, the user's default folder for documents is displayed (on all OS)
- [ ] When opening the upload files modal subsequent times, the previous folder is displayed (afaik this is current behaviour)

**Background**
Currently on Windows, it opens by default to the Hypergraph installation folder

juliangruber:
We don't have more control than telling the OS the default path to use


jameslibscie:
@juliangruber It opens to Documents every time now 👍 This is definitely an improvement over how it previously was, but there must be a way to remember the previous location, right? If that's also a very quick fix, perhaps we could implement that now - otherwise park it for later?

juliangruber:
> This is definitely an improvement over how it previously was, but there must be a way to remember the previous location, right? If that's also a very quick fix, perhaps we could implement that now - otherwise park it for later?

Electron doesn't make this possible for us, no.

PR #93, opened 2020-05-27
fix profile title scrolling. closes #86

### LINKED ISSUE #86: Scrolldown of header bar issue
When I have more content than the profile screen can display at once, I scroll down and get the following visual bug

Screenshot 2020-05-26 at 10 29 05

I was running Hypergraph 0.2.0 on the latest macOS Catalina.

juliangruber:
See https://github.com/hypergraph-xyz/desktop/issues/86#issuecomment-634481943 for discussion.


jameslibscie:
Hmm, this does make the editing feel a lot less smooth. The transition to the yellow/orange bars is also lost right? Would it be possible to do a transition to the new state? Whereby the content quickly fades out / fades in and the yellow bars pulsate like they did before?

juliangruber:
You're right, I missed that this removed the transition effect. Since we're redrawing the whole DOM to add the `<form>` wrapper, CSS transitions don't work.

I see these paths:

- remove the sticky scrolling
- use one `<form>` element per input, which makes it more tricky

I'll see if the multiple `<form>` element solution becomes horrible or not

juliangruber:
Ok I got back the transitions. @jameslibscie please review again

jameslibscie:
The scrolling is fixed and editing, but for some reason the bottom border for the menu items seems to have disappeared:
image
My local copy of master from earlier today has the border.

jameslibscie:
Do you mean @psobrakseaton that the description area would also become sticky while scrolling?

psobrakseaton:
> Do you mean @psobrakseaton that the description area would also become sticky while scrolling?

Basically, no movement from the header bar or the profile area while you are in edit mode. No scrolling past them or collapsing while editing.

If changes have been saved or cancelled and you are scrolling though the content modules, that area can collapse.

juliangruber:
> The scrolling is fixed and editing, but for some reason the bottom border for the menu items seems to have disappeared:

Oops yes I needed to redo some button components because of this change and missed that. Will fix.

juliangruber:
> Basically, no movement from the header bar or the profile area while you are in edit mode. No scrolling past them or collapsing while editing.
> 
> If changes have been saved or cancelled and you are scrolling though the content modules, that area can collapse.

Let me try that, however my intuition says that these context specific changes are confusing both to the user and will make the implementation also more confusing.

psobrakseaton:
> > Basically, no movement from the header bar or the profile area while you are in edit mode. No scrolling past them or collapsing while editing.
> > If changes have been saved or cancelled and you are scrolling though the content modules, that area can collapse.
> 
> Let me try that, however my intuition says that these context specific changes are confusing both to the user and will make the implementation also more confusing.

Yeah, I really trust your intuition. If we could try it that'd be great, but your gut feelings about these things haven't led us wrong in the past. Let me think on some of the other solutions as well.

juliangruber:
> Yeah, I really trust your intuition. If we could try it that'd be great, but your gut feelings about these things haven't led us wrong in the past. Let me think on some of the other solutions as well.

I tried that and we're not going to do it. That just becomes very messy!

psobrakseaton:
Ok, lets scrap it then. I'll take a look at the latest build...

juliangruber:
> Ok, lets scrap it then. I'll take a look at the latest build...

Did you have a look?

PR #94, opened 2020-05-27
add dynamic page titles. closes #83

### LINKED ISSUE #83: dynamic page titles
Using react-router we can change the page title as the user navigates. This will make the navigation data in Chatra more useful:

Screenshot 2020-05-25 at 12 13 17

juliangruber:


juliangruber:
@jameslibscie I had to disable changing the title on Windows, however on mac and linux it will change it still.

jameslibscie:
Hmm, could you show me what that looks like on Mac and Linux? And is it basically impossible to change the page title on those platforms while keeping the window title?

jameslibscie:
Ah, I may have misunderstood you. By title, do you mean the window title or the page title?

juliangruber:
Dang I realized there actually is a dynamic window title on mac too. We'll have to close this :(

jameslibscie:
@juliangruber The purpose of this was to make sure Chatra shows on what page the user is, right? Is there perhaps a way to pass a custom title / page / something to Chatra instead of using the page title mechanism?

juliangruber:
They only have a [`pageView` method](https://chatra.com/help/api/#pageview) which will read the title off `document.title`. We could try setting the title, calling that method, then resetting it. Hopefully no flashing would be visible.

jameslibscie:
Cool! It feels a bit out-of-scope for this milestone, so we can park this unless you're enthusiastic about getting this solved :)

PR #99, opened 2020-06-04
Add global AddContent button

juliangruber:


juliangruber:
> Solution 1 - Disable Button

How is the button disabled in this example?

> Solution 2 - Replace form back button

There's no form or back button in this screen?

juliangruber:
Oh I see, your example isn't on the first page.

> Solution 1 - Disable Button

This makes me think whether the add content form should actually be an overlay?

> Solution 2 - Replace form back button

This is very confusing. Having a back button in the menu bar like this is very unexpected.

jameslibscie:
@juliangruber If you click the button in the bottom-left, it takes you to 2 variations of the add content page.

The text on the Workspace will need updating to match the new location. Perhaps:

> Click the <hex add content icon> button in the bottom-left corner to get started 👈

Or perhaps without the emoji. Saying 'Add content' instead of using the icon is also fine.

Regarding the button when the Add content page is open, I think it should be disabled. Turning it into a back button breaks with the rest of the interface too much.

jameslibscie:
@juliangruber I see what you mean about an overlay, as technically all the menu items should really be disabled. It also feels to me like this is good enough for now and perhaps when we *need* an overlay for something else, we can design that and then migrate the add content to an overlay.

juliangruber:
> The text on the Workspace will need updating to match the new location. Perhaps:
> 
> > Click the  button in the bottom-left corner to get started 👈
> 
> Or perhaps without the emoji. Saying 'Add content' instead of using the icon is also fine.

Good idea, what about inlining the icon? "Click <IC+N> to add content"

juliangruber:
> @juliangruber I see what you mean about an overlay, as technically all the menu items should really be disabled. It also feels to me like this is good enough for now and perhaps when we _need_ an overlay for something else, we can design that and then migrate the add content to an overlay.

I disagree that all the menu items need to be enabled in this screen, it's just that if you use them we need to warn the user about losing their data. I actually think there doesn't need to be a solution to this button at all, imo it's fine as it is right now.

jameslibscie:
Inlining the icon is actually what I suggested, but GitHub seems to have removed / hidden it because it was in between <>. So great if we can do that :)

psobrakseaton:
OK, so replied via email and it doesn't appear to have updated here.

So I read your discussion, and I am not sure entirely where we stand at this moment.

Are we running with disabling it at the moment or would you like the overlay concept?

psobrakseaton:
> > @juliangruber  I actually think there doesn't need to be a solution to this button at all, imo it's fine as it is right now.

I am definitely bothered by the current implementation. The empty button just feels off to me.

juliangruber:
> Are we running with disabling it at the moment or would you like the overlay concept?

There's no decision yet. To summarize: James wants to disable the button, I want to either keep it as is or change the form to an overlay.

It doesn't make sense to me to disable menu bar items, when we already have a pattern for it: The link to page you're already on will be marked as active (and not disabled).

jameslibscie:
I would personally vote for disabling it now, but only if it's little effort. Then at a later stage, we could consider moving to an overlay, especially if we need an overlay for some other feature that would apply well to the add content concept.

The only thing that *needs* changing now imo is the text on the Workspace.

jameslibscie:
@juliangruber Interesting, hadn't thought of that yet. Making it "active" like other menu items would also make sense to me.

juliangruber:
If the problem is that if we mark the AddContent button as active (just like the Workspace and Profile links above) it becomes too bold, then maybe we need a different active style for it, or a different style where the matching active style isn't too bold.

jameslibscie:
I think it's OK to treat the add content button differently from other menu items and therefore opt for disable rather than active. It's visually the calmest and imo doesn't break the current pattern too hard as you could see the add content button as its own entity.

psobrakseaton:
Thanks for the summation, Julian. I am prepping a couple an idea right now for al alternative to the active state. Give me a moment...

psobrakseaton:
[Ok, here's a suggestion for the active states ](https://www.figma.com/proto/hKyO42soxOl4TIiIfqkgzG/%CE%B2-global-add-content-button?node-id=89%3A159&viewport=509%2C567%2C0.4696509838104248&scaling=scale-down)

Thoughts?

I talked with Julian yesterday about some of the problems I was running into with some of the button styling and I really think this is another area where the design needs to evolve, and I am going to try to fast track that next milestone.

psobrakseaton:
I tried some other things that just didn't work, so I think having an "active" color will be best for now.

jameslibscie:
I like it, but have a couple of questions:
1. How is this accessibility-wise? Is the font colour enough of an indication?
2. Does the hover state then also change to match the active state?
3. If not, then we should disable the hover state on active menu items, right?

juliangruber:
> * How is this accessibility-wise? Is the font colour enough of an indication?

I kind of agree, purple text on black background is somewhat not easy to read.

> * Does the hover state then also change to match the active state?
> * If not, then we should disable the hover state on active menu items, right?

Active menu items already don't have a hover state :)

juliangruber:
> Thoughts?

I think this might be better than before, could you add the missing hover states to the proto so we can fully evaluate this proposal?

psobrakseaton:
Yup, I definitely think there is a contrast problem.

I think the best way for this to proceed is to go with this "good enough" state we have right now and I will make more prototypes we can all assess together in the near future.

Making notes about this for milestone 4...

juliangruber:
Screenshot 2020-06-08 at 12 38 06

jameslibscie:
Really nice with the inline icon, but I kinda miss the emoji. Maybe we could add a ✏, 📝, ✍🏻  or something more abstract like 🚀 or 🌟. I'll approve anyway, so feel free to do this at your own discretion :)

juliangruber:
> Really nice with the inline icon, but I kinda miss the emoji. Maybe we could add a ✏, 📝, ✍🏻 or something more abstract like 🚀 or 🌟. I'll approve anyway, so feel free to do this at your own discretion :)

I don't miss the emoji tbh 🤷‍♂️ I'm going to merge for now and we can add an issue if we miss it

PR #104, opened 2020-06-11
Update/workspace drafts

### LINKED ISSUE #103: Rename "workspace" to "drafts"
**User story**
As a content creator
I want there to be a clear distinction between my drafts and my profile
so that I can get a better overview of my in-progress work

**Acceptance criteria**

- [x] Workspace is renamed to Drafts
- [x] Filter Drafts page to only display draft content
- [x] Replace "Go to workspace/drafts" text with "Get started!" (or similar)
- [x] If there is listed content, but no draft content, replace "Nothing here yet" text with "No drafts! All your work is now on your profile 😎" (or similar)

jameslibscie:
At the end of the welcome modals, the "Go to workspace" becoming "Go to drafts" didn't sit that well with me, so I changed it to "Get started!"

I also added a text for when there are no drafts but there is content on the profile.

Couple of code things I'm not sure about:
- Logic for determining whether a module is listed or not is now duplicated across `drafts.js` and `module.js`. I couldn't find a pattern for dealing with this. Any ideas @juliangruber? Normally I would have this as a method of a `Module` class. Would a static method on `Module` make sense here?
- I nested / chained ternaries - anything against that?

Closes #103


jameslibscie:
I've added a `Module.isListed(mod, profile)` static method. Curious what you think of that approach

app/components/drafts/drafts.js
```diff
@@ -13,19 +13,24 @@ const StyledAddContent = styled(AddContent)`
...
 
-export default ({ p2p }) => {
+export default ({ p2p, profile }) => {
   const [modules, setModules] = useState()
+  const [hasListedContent, setHasListedContent] = useState()
 
   useEffect(() => {
     ;(async () => {
-      setModules(await p2p.listContent())
+      const allContent = await p2p.listContent()
```
juliangruber, line 22: for consistency with other components this should be renamed to `modules`.

app/components/drafts/drafts.js
```diff
@@ -13,19 +13,24 @@ const StyledAddContent = styled(AddContent)`
...
+      const writableContent = allContent.filter(content => content.metadata.isWritable)
```
juliangruber, line 23: we don't differentiate between writable and unwritable content yet, as there's no way to get unwritable content into your app

app/components/drafts/drafts.js
```diff
@@ -13,19 +13,24 @@ const StyledAddContent = styled(AddContent)`
...
+      const draftContent = writableContent.filter(mod => !Module.isListed(mod, profile))
```
juliangruber, line 24: can we rename this to `drafts`? That's shorter and doesn't have the singular vs plural problem of the word `content`.

app/components/drafts/drafts.js
```diff
@@ -44,6 +49,10 @@ export default ({ p2p }) => {
               <>
                 {modules.length ? (
                   'You’ve reached the end! ✌️'
+                ) : hasListedContent ? (
```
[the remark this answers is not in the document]
juliangruber: Good idea to add this footer text state!

app/components/menu/menu.js
```diff
@@ -82,7 +82,7 @@ const Menu = () => {
       <StyledLogo onClick={() => history.push('/')} />
       <StyledRow>
         <StyledNavLink to='/' isActive={(_, location) => isWorkbench(location)}>
```
juliangruber, line 84: could you also rename the `isWorkbench` check (which as you can see I forgot to rename before already :D)?

app/components/module/module.js
```diff
@@ -216,4 +206,18 @@ const Module = ({ p2p, mod, pad, to, isParent }) => {
   )
 }
 
+Module.isListed = (mod, profile) => {
```
juliangruber, line 209: `Module` here isn't just a class, it's a react component. This functionality is better stored in `app/lib`, where pure / react independent functions live. Also `Module` is a somewhat confusing name because `module` is a node.js global. What about for now moving this single check to `app/lib/is-module-listed.js`?

app/components/drafts/drafts.js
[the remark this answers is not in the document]
jameslibscie: The variable name did indeed mean that the user has one or more listed content modules. So it would only work as an "else" condition in this case. Changed to `hasOnlyListedContent` as that's clearer and less likely to break if we switch things around :)

app/components/drafts/drafts.js
jameslibscie: Done!

app/components/drafts/drafts.js
jameslibscie: Was wondering about this -> removed!

app/components/drafts/drafts.js
jameslibscie: Done!

app/components/menu/menu.js
jameslibscie: Done!

app/components/module/module.js
jameslibscie: It works to add a static method to a function, although I can imagine that's an odd pattern. Moved to lib :)

app/components/drafts/drafts.js
```diff
@@ -49,7 +49,7 @@ export default ({ p2p, profile }) => {
               <>
                 {modules.length ? (
                   'You’ve reached the end! ✌️'
-                ) : hasListedContent ? (
+                ) : hasOnlyListedContent ? (
```
juliangruber, line 52: I see now why you named that variable `hasListedContent` originally, really it is `hasListedAndOnlyListedContent`, right?

```suggestion
                ) : profile.rawJSON.contents.length && hasOnlyListedContent ? (
```

app/components/drafts/drafts.js
```diff
@@ -13,23 +14,27 @@ const StyledAddContent = styled(AddContent)`
...
+export default ({ p2p, profile }) => {
+  const [drafts, setDrafts] = useState()
+  const [hasListedContent, setHasListedContent] = useState()
 
   useEffect(() => {
     ;(async () => {
-      setModules(await p2p.listContent())
+      const modules = await p2p.listContent()
+      const drafts = modules.filter(mod => !isModuleListed(mod, profile))
+      setHasListedContent(profile.rawJSON.contents.length > 0)
```
juliangruber, line 25:
```suggestion
      setHasListedContent(profile.rawJSON.contents.length > 0)
      const modules = await p2p.listContent()
      const drafts = modules.filter(mod => !isModuleListed(mod, profile))
```

reordered because the `setHasListedContent` call has nothing to do with the two lines above

app/components/drafts/drafts.js
```diff
@@ -13,23 +14,27 @@ const StyledAddContent = styled(AddContent)`
   top: -2px;
 `
 
-export default ({ p2p }) => {
-  const [modules, setModules] = useState()
+export default ({ p2p, profile }) => {
+  const [drafts, setDrafts] = useState()
+  const [hasListedContent, setHasListedContent] = useState()
```
juliangruber, line 19:
```suggestion
  const hasListedContent = profile.rawJSON.contents.length > 0
```

i forgot how to think, even better this way 😅

PR #138, opened 2020-07-08
Change protocol from dat:// to hyper://

chartgerink:
I noticed this was still outdated, and given the recent change to the specs this is an important tidbit :)


app/components/create/create.js
```diff
@@ -102,7 +102,7 @@ const Create = ({ p2p, profile }) => {
               title,
               description,
               authors: [profile.rawJSON.url],
-              ...(parentUrl && { parents: [`dat://${parentUrl}`] })
+              ...(parentUrl && { parents: [`hyper://${parentUrl}`] })
```
jameslibscie, line 105: The protocol should actually be removed here to be up to date with the specs. We only use the protocol in the URL and now ban it for all other fields for simplicity.

The SDK should allow parents here without the protocol already, but will disallow the protocol in a later version.

jameslibscie: See also https://github.com/p2pcommons/sdk-js/issues/179

jameslibscie:
Still waiting for https://github.com/p2pcommons/sdk-js/pull/189

PR #141, opened 2020-07-13
Share profile. Closes #108

### LINKED ISSUE #108: Share profile
https://www.figma.com/file/BQwYpFpsOJaxpqmUOVTRSH/%CE%B2---networking-feature-share-profile?node-id=61%3A1534

- [x] add share button to profile
- [x] open modal with unversioned profile link
- [ ] add copy to clipboard button
- [x] close modal

juliangruber:
Screenshot 2020-07-13 at 17 03 36

Can we ship this without the "copy to clipboard" button for now? Maybe we have time before this release, but this works without it just as fine and saves some time.


juliangruber:
> Any form of sharing is always non-reversible in the sense that the other may make a copy for personal use.

Not correct, I can share a blu-ray with you but you can't make a copy without violating the law. I think this is good to mention, especially for P2P tech, where you make a copy _by default_.



> Also, the copy will need expanding once we have a way to view remote profiles, because this also enables sharing remote profiles. I think this should be different copy, displayed conditionally, and can help with the copy for the share modal for other people's profiles.

@psobrakseaton and me talked about this yesterday and he wanted to bring it up with you :)

jameslibscie:
> > Any form of sharing is always non-reversible in the sense that the other may make a copy for personal use.
> 
> Not correct, I can share a blu-ray with you but you can't make a copy without violating the law. I think this is good to mention, especially for P2P tech, where you make a copy _by default_.
> 
> > Also, the copy will need expanding once we have a way to view remote profiles, because this also enables sharing remote profiles. I think this should be different copy, displayed conditionally, and can help with the copy for the share modal for other people's profiles.
> 
> @psobrakseaton and me talked about this yesterday and he wanted to bring it up with you :)

True. What I mean is - the type of sharing that people are used to. I do agree though that the "copy by default" part is different from non-P2P applications. I'm curious to find out what users think of this.

jameslibscie:
@juliangruber @psobrakseaton I've noticed that other fields now also get an "active" inner border, for example during profile creation and on the add content page. Is that intentional?

juliangruber:
> @juliangruber @psobrakseaton I've noticed that other fields now also get an "active" inner border, for example during profile creation and on the add content page. Is that intentional?

This is because of the recent electron update - Chrome changed its default styles. I'll take a look.

PR #147, opened 2020-07-17
Add database backup & restore. Closes #114

### LINKED ISSUE #114: Back up & restore
Add these features to the menu bar:

- [x] Back up to zip / folder
- [x] Restore from zip / folder

If necessary, also set up the menu bar in the first place.

juliangruber:
image

This backs up and restores `~/.p2pcommons` using `.zip` files. This feels great to have, especially during testing!


chartgerink:
Does this also back up and restore private keys (from `~/.hyperdrive`)?

juliangruber:
It does not. I wasn't aware of this folder, and thought everything would be stored in `~/.p2pcommons`. @dpaez can you confirm backing up these two folders should work?

dpaez:
Hi all 👋, We are not using `.hyperdrive` directory. 
For now, saving `~/.p2pcommons` it's enough. If we use the daemon,  then the story would be a bit different, but that is not the case.

jameslibscie:
I've tested this out for moving keys & data from machine to machine and it works:
- Created profile & content on desktop
- Backup database
- Reset database
- Create new profile
- Import database on laptop
- Edit profile & content on laptop
- Access profile from desktop
- Updated profile & content shows correctly

Issue #156, opened 2020-07-22
Error upon npm run start


chartgerink:
## Description

I'm trying to run the `main` branch as a way of testing where another issue might be coming from (see context).

## To Reproduce

```
git clone git@github.com:hypergraph-xyz/desktop
cd desktop
npm install
npm run build
npm run start
```

## Expected behavior

Boots Hypergraph

## Actual behavior

Returns error

```
(node:10738) UnhandledPromiseRejectionWarning: Error: Invalid header: Does not start with Cr24
    at crxToZip (/Users/chjh/git/hypergraph-xyz/desktop/node_modules/unzip-crx/dist/index.js:32:15)
    at /Users/chjh/git/hypergraph-xyz/desktop/node_modules/unzip-crx/dist/index.js:57:32
    at callHanler (/Users/chjh/git/hypergraph-xyz/desktop/node_modules/yaku/lib/yaku.js:719:16)
    at tryCatcher (/Users/chjh/git/hypergraph-xyz/desktop/node_modules/yaku/lib/yaku.js:436:32)
    at /Users/chjh/git/hypergraph-xyz/desktop/node_modules/yaku/lib/yaku.js:575:38
    at flush (/Users/chjh/git/hypergraph-xyz/desktop/node_modules/yaku/lib/yaku.js:478:17)
    at processTicksAndRejections (internal/process/task_queues.js:76:11)
(node:10738) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:10738) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

## System

- OS (incl. version): macOS 10.15.6
- Hypergraph Desktop version: `main`
- Node v12.18.2
- npm 6.14.5

## Additional context

I was trying this to debug another issue where the Hypergraph window doesn't open after booting the app. This is all being run on a clean install of macOS. When I run `npm test` it passes without a hitch.

jameslibscie:
I use the command `npm start` rather than `npm run build` and `npm run start`. Does that work for you?

chartgerink:
Same error, but thanks for the suggestion 😸

juliangruber:
@chartgerink does it still not launch for you?

juliangruber:
[...]

PR #157, opened 2020-07-22
Delete module folders from disk

chartgerink:
The module folders were not deleted from disk when the `Delete content` button was pressed. I talked to @dpaez who mentioned there is a hard delete option that is not defaulted to. In this PR, I added that option

Note: I ran `npm test` successfully, but I could not test the functionality because of #156


app/components/content/content.js
```diff
@@ -299,7 +299,7 @@ const Content = ({ p2p, content, renderRow }) => {
               isLoading={isDeleting}
               onClick={async () => {
                 setIsDeleting(true)
-                await p2p.delete(content.rawJSON.url)
+                await p2p.delete(content.rawJSON.url, true)
```
juliangruber, line 302:
```suggestion
                const deleteFiles = true
                await p2p.delete(content.rawJSON.url, deleteFiles)
```

Unfortunately this API uses boolean arguments, which need some documentation

jameslibscie:
I think for this to be safe, there needs to be a warning dialog for the user to confirm deletion. Right now, the user can misclick and lose data from Hypergraph, but at least the files will still be there (if they can find them). This is also not ideal, but losing the files as well is even worse.

Basically the delete button was never spec'd out and built based on Patch's designs.

My suggestion would be to add a work item for perhaps the 2nd half of this cycle if we deem it important enough that encompasses this warning dialog and file deletion. The warning dialog would require design work.

chartgerink:
@jameslibscie how about we use a simple dialog box for now? I agree a confirmation is an important bit to prevent accidental deletes 😊

jameslibscie:
> @jameslibscie how about we use a simple dialog box for now? I agree a confirmation is an important bit to prevent accidental deletes 😊

I'd be fine with that. Would @psobrakseaton have an objection to doing this for now and replacing it with a styled dialog box later?

juliangruber:
It should look something like this:

```js
import { remote } from 'electron'
const { response } = await remote.dialog.showMessageBox(remote.getCurrentWindow(), {
  type: 'warning',
  buttons: ['Delete all files', 'Cancel'],
  message: 'Do you really wanna?'
})
if (response === 1) return
```

jameslibscie:
Suggestion for copy:
> This will also delete these files from your Hypergraph folder. Are you sure you want to delete this content?
> Delete files / cancel

jameslibscie:
@chartgerink Are you still aware of this PR? Waiting for something perhaps?

chartgerink:
I am aware of it but was (1) unsure whether the expectation was for me to implement because of my limited implementation skills and (2) if so was blocked by #156. Given that #195 fixes it I just tried something out.

```js
              onClick={async () => {
                setIsDeleting(true)
                const deleteFiles = true
                const { response } = await remote.dialog.showMessageBox(remote.getCurrentWindow(), {
                  type: 'warning',
                  buttons: ['Delete all files', 'Cancel'],
                  message: 'Do you really wanna?'
                })

                if (response === 1) {
                  await p2p.delete(content.rawJSON.url, deleteFiles)
                  history.push('/')
                }
              }}
```

It gives me the warning modal but it doesn't delete. If somebody can help we can make this change rather easily yes :)

juliangruber:
change it to `if (response === 0)` :)

juliangruber:
@chartgerink do you want me to take this over?

chartgerink:
Thanks @juliangruber! Pushed the fix. I love contributing tidbits where I can so I was happy to be able to do this with your help 🙌

app/components/content/content.js
```diff
@@ -321,8 +321,20 @@ const Content = ({ p2p, content, renderRow }) => {
...
                 setIsDeleting(true)
-                await p2p.delete(content.rawJSON.url)
-                history.push('/')
+                const deleteFiles = true
+                const { response } = await remote.dialog.showMessageBox(
+                  remote.getCurrentWindow(),
+                  {
+                    type: 'warning',
+                    buttons: ['Delete all files', 'Cancel'],
+                    message: 'Do you really wanna?'
```
juliangruber, line 330: This copy sounds unprofessional to me. What about
```suggestion
                    message: 'Would you also like to remove all of this module\'s files from disk? This action cannot be undone.'
```

app/components/content/content.js
```diff
@@ -321,8 +321,20 @@ const Content = ({ p2p, content, renderRow }) => {
...
+                  }
+                )
+
+                if (response === 0) {
```
juliangruber, line 334: In order to follow existing patterns in other components

```suggestion
                if (response === 1) return

```

app/components/content/content.js
```diff
@@ -321,8 +321,20 @@ const Content = ({ p2p, content, renderRow }) => {
...
+                  await p2p.delete(content.rawJSON.url, deleteFiles)
```
juliangruber, line 335: could you move the `deleteFiles` variable declaration right above this line? This way it is in context

app/components/content/content.js
jameslibscie: The user doesn't get the choice with this modal between deleting the files or just deleting the module - it's either all or nothing. Delete *all* files also sounds unnecessarily dramatic to me. I made this suggestion above:

> This will also delete these files from your Hypergraph folder. Are you sure you want to delete this content?
> Delete files / cancel

chartgerink: 😂 ohh hahaha i totally focused on the code i didn't think about the copy of course! 

Suggested copy looks good 👍

app/components/content/content.js
chartgerink: Didn't know this trick :) Thanks!

juliangruber:
gif drop it like it's hot

PR #184, opened 2020-07-24
Add sort content modules. Closes #163

### LINKED ISSUE #163: Sort content modules
Sort all content lists new-old

juliangruber:
In the drafts and the profile screen, sort content by date descending.


juliangruber:
@jameslibscie when you register a module, that will change its last modification time. Say you have two modules, called `old` and `young`, which conveniently are old and young. If you now register `young`, then `old`, `old` will have the newer `lastModified` time, and appear on top of the list.

In the video you showed this doesn't happen perfectly, as the modules `CHI` and `CHI2` are out of order, however it's correct that `windows.zip` would be at the bottom of the list, since it was registered first and thus is the oldest.

jameslibscie:
> @jameslibscie when you register a module, that will change its last modification time. Say you have two modules, called `old` and `young`, which conveniently are old and young. If you now register `young`, then `old`, `old` will have the newer `lastModified` time, and appear on top of the list.
> 
> In the video you showed this doesn't happen perfectly, as the modules `CHI` and `CHI2` are out of order, however it's correct that `windows.zip` would be at the bottom of the list, since it was registered first and thus is the oldest.

Afaik the order in which I registered the modules didn't have an effect on their display order. I'm also not sure that it's correct that the `lastModified` would be updated, as registering a module should only write to the profile. The content itself doesn't need to change at all.

juliangruber:
> Afaik the order in which I registered the modules didn't have an effect on their display order. I'm also not sure that it's correct that the `lastModified` would be updated, as registering a module should only write to the profile. The content itself doesn't need to change at all.

Oh hmm you're right. I'll take a closer look

juliangruber:
I've created three modules, this is their `lastModified` dates:

```
b-Tue Jul 28 2020 15:47:30 GMT+0200 (Central European Summer Time)
c-Tue Jul 28 2020 15:47:30 GMT+0200 (Central European Summer Time)
a-Tue Jul 28 2020 15:47:30 GMT+0200 (Central European Summer Time)
```

I created a first, then b, then c. I'm not sure why the date is the same for all :/

juliangruber:
I'm going to merge this here, can you please check this with the SDK side?

PR #186, opened 2020-07-24
Add share content. Closes #112

juliangruber:
image

## TODO

- [x] final copy @jameslibscie 
- [x] add ability to choose versioned vs unversioned link

PR #187, opened 2020-07-24
Add view content by link. Closes #113

### LINKED ISSUE #112: Share content
- [ ] add share button to content
- [ ] open modal with versioned and/or unversioned content link
- [ ] add copy to clipboard button(s)
- [ ] close modal

[Figma](https://www.figma.com/file/tAK0AcupI5t8B8NKwBflaq/%CE%94-networking-feature-share-content?node-id=87%3A486)

### LINKED ISSUE #113: View content by link
- [x] add search button
- [x] open modal with search input
- [x] load content
- [x] display content

juliangruber:
This is based on https://github.com/hypergraph-xyz/desktop/pull/186.

## TODO

- [x] Find modal text needs to be adjusted. There is now the case where you share content, not a profile. And also the content doesn't have to be yours.

image


juliangruber:
> The navigation is slightly broken in the sense that if you view a newer version and then an older version (probably the same the other way round) it doesn't navigate to the older version. If you view a newer version, then navigate away, then view the older version it works fine. Seems like something to fix at a later stage as this is probably quite an uncommon use case in the beginning.

fixed :)

PR #190, opened 2020-07-24
Add change parent (one parent only). Closes #159

### LINKED ISSUE #159: Change parent (one parent only)
A simple dropdown is enough.

- [ ] add parents dropdown
- [ ] show only registered contents

Review: @chartgerink @jameslibscie

juliangruber:
image

This was created without designs.


chartgerink:
Looking really good :) I +1 what @jameslibscie said about the dropdown; it's a component that reoccurs and can benefit from some more design work in another cycle.

I only saw encountered one issue as I installed this version: I could only select content modules I registered in this version, but not ones that were already present. Attached a screen showing my profile with registered modules and subsequently a screen showing my options.

Screenshot 2020-07-24 at 17 53 38
Screenshot 2020-07-24 at 17 53 46

Note that opening the content module registered in the old version doesn't display as registered in the new version
Screenshot 2020-07-24 at 17 57 45

This might be a different issue, but thought it was worthwhile to note.

juliangruber:
> I only saw encountered one issue as I installed this version: I could only select content modules I registered in this version, but not ones that were already present. Attached a screen showing my profile with registered modules and subsequently a screen showing my options.

Hm well the current logic is that potential parents are found by iterating over all modules and checking which has been registered in its latest version. Instead I think we can iterate over all profiles and then list the modules they have registered. If they registered the same content twice, I guess we could list `name (v13)` and `name (v18)`?

chartgerink:
> > I only saw encountered one issue as I installed this version: I could only select content modules I registered in this version, but not ones that were already present. Attached a screen showing my profile with registered modules and subsequently a screen showing my options.
> 
> Hm well the current logic is that potential parents are found by iterating over all modules and checking which has been registered in its latest version. Instead I think we can iterate over all profiles and then list the modules they have registered. If they registered the same content twice, I guess we could list `name (v13)` and `name (v18)`?

Multiple registrations we can tackle another time? My main worry is that people can't select registrations from Hypergraph `v0.3.0` (for example) in `v0.4.0`, which happened to me here. Regardless, I didn't systematically test this so might be just an off-case too.

juliangruber:
> This may omit the content shared/added to the database that's not captured by the profiles? (like an orphan module so to speak) If so, I guess we'll see whether that's an issue at a later time.

You're right! I'll add that too

juliangruber:
> This may omit the content shared/added to the database that's not captured by the profiles? (like an orphan module so to speak) If so, I guess we'll see whether that's an issue at a later time.

With orphaned modules it's currently hard to test if they're valid, since we don't have their author's profile available. Let's indeed add this later.

PR #206, opened 2020-07-30
Add Mini Map. Closes #169

### LINKED ISSUE #169: View parents context for a content module (mini-mapping)
Show a specific minimapping icon depending on the number of parent modules.

[Figma](https://www.figma.com/file/FM8tovSvzNVCKqqOKdseaq/%23169---View-parents-context-for-a-content-module-(mini-mapping)) | Review @chartgerink @psobrakseaton

juliangruber:
image

image

Closes #169

## TODO

- [x] the SVG draws some unwanted white borders
- [x] should the map be left or right aligned?
- [x] what's the best behavior on click?


jameslibscie:
I think it should go to the parent when clicked, but I expect that this is going to be confusing if there's no tooltip.

juliangruber:
> I think it should go to the parent when clicked, but I expect that this is going to be confusing if there's no tooltip.

I checked in with Patch, it should toggle the follows from list. This way we can remove the clumsy `Follows from v` button.

jameslibscie:
> > I think it should go to the parent when clicked, but I expect that this is going to be confusing if there's no tooltip.
> 
> I checked in with Patch, it should toggle the follows from list. This way we can remove the clumsy `Follows from v` button.

Even better!

psobrakseaton:
> > I think it should go to the parent when clicked, but I expect that this is going to be confusing if there's no tooltip.
> 
> I checked in with Patch, it should toggle the follows from list. This way we can remove the clumsy `Follows from v` button.

For now it should toggle the follows from list. And I am in favor of removing the "Follows from" text. It is awkward and we implemented it really fast. I am keep a running list of things that will eventually need to change with this design (hover tooltip, multiple parents, follows from page, etc.). Please keep notes if you find anything amiss.

juliangruber:
I just realized that without the Follows From link nothing explains what the minimap means. I think we should add it back.

jameslibscie:
> I just realized that without the Follows From link nothing explains what the minimap means. I think we should add it back.

Would having this covered in the onboarding tour be enough do you think?

psobrakseaton:
> I just realized that without the Follows From link nothing explains what the minimap means. I think we should add it back.

Yup, I was just writing this. This task is dredging up a lot of questions that we need to consider. Can we table this until next week?

juliangruber:
> Would having this covered in the onboarding tour be enough do you think?

I don't think you can learn and remember all of the app's concepts at once

psobrakseaton:
> > Would having this covered in the onboarding tour be enough do you think?
> 
> I don't think you can learn and remember all of the app's concepts at once

Agreed. My current thinking is that the "Follows" from should be moved to the dropdown in some way so it can repeat to the next item as well. I would also like to add some sort of "thread" graphic like this to the dropdown.
Screen Shot 2020-07-30 at 13 10 32

It is a very well-known and established pattern and should really be used.

I think this task has jumped the gun a tiny bit and I'd like to move it back to review if you are all in agreement.

jameslibscie:
> > Would having this covered in the onboarding tour be enough do you think?
> 
> I don't think you can learn and remember all of the app's concepts at once

True dat.

I would be very much in favour of having a tooltip displaying something like:
> Follows from:
> <Title>, <Authors>
And have that expand the chain of parents.

jameslibscie:
> I think this task has jumped the gun a tiny bit and I'd like to move it back to review if you are all in agreement.

Sounds good to me!

juliangruber:
> I think this task has jumped the gun a tiny bit and I'd like to move it back to review if you are all in agreement.

What do you mean by this? It's already in review

psobrakseaton:
> > I think this task has jumped the gun a tiny bit and I'd like to move it back to review if you are all in agreement.
> 
> What do you mean by this? It's already in review

Sorry, I meant that there are a bunch of questions that have come up and I don't think it is fully baked yet. I'm going to revisit this task on Monday since I'm not working tomorrow.

jameslibscie (approved):

chartgerink (approved):

psobrakseaton (approved):
Approved!

PR #208, opened 2020-08-04
Update welcome modals

jameslibscie:
- vault text expanded (people in tests really weren't getting this without the p2p context)
- libscie mention on first page + dedicated libscie page
- explained profile name (this also caused some small confusion often)
- commas and periods
- updated `Anchor` style to reduce distance to underline so that it works better in paragraphs
- updated `Anchor` so that `href` automatically opens externally for links


app/components/anchor.js
```diff
@@ -7,7 +8,7 @@ const StyledAnchor = styled.a`
   color: ${props => (props.disabled ? gray : white)};
   border-bottom: 2px solid
     ${props => (props.disabled ? gray : props.color || purple)};
-  display: inline-block;
+  display: inline;
```
juliangruber, line 11:
```suggestion
```

This is already the default for `<a>` elements

app/components/anchor.js
```diff
@@ -23,10 +24,23 @@ const StyledAnchor = styled.a`
...
-const Anchor = ({ onClick, disabled, ...props }) => (
+const Anchor = ({ onClick, disabled, href, ...props }) => (
   <StyledAnchor
     disabled={disabled}
-    onClick={() => !disabled && onClick && onClick()}
+    href={href}
+    onClick={e => {
+      if (disabled) {
+        return
+      }
```
juliangruber, line 34:
```suggestion
      if (disabled) return
```

terser

jameslibscie:
@psobrakseaton This has an effect on the Liberate Science link on the Feed and other pages too. Could you check whether the change is ok? It brings the underline closer to the text a little.

app/components/welcome/welcome.js
```diff
@@ -156,20 +163,75 @@ const dialogs = [
...
+          <IllustrationVault />
+        </Illustration>
+        <Heading>Introducing the Vault</Heading>
+        <p>
+          We use peer-to-peer technology: Hypergraph is powered by the people using it!
+          Share your content as you'd like, without anybody getting in between.
+        </p>
+        <p>
+          Hypergraph Vault makes your work available even when your computer is offline. Plus, we're
+          working with libraries on archiving it for future generations! 👵👨👶
```
chartgerink, line 180:
```suggestion
          working with libraries on archiving it for future generations! 👵🏾👴🏻👶
```

My only comment on this is for a bit more diversity in the emojis. @lhehnke what do you think? 

Screenshot 2020-08-04 at 18 03 26

Otherwise looks good to me and I'll approve once this is discussed :)

app/components/welcome/welcome.js
```diff
@@ -156,20 +163,75 @@ const dialogs = [
...
+        </p>
+        <p>
+          Until January 1st, 2021, Hypergraph Vault is free, while we figure out the costs.
+          We promise to keep it to one-time fees for lifetime service.
```
chartgerink, line 184:
```suggestion
          It won't be more than five coffees ☕ ☕ ☕ ☕ ☕ 
```

On second thought, lets scrap the promise so we don't end up reneging. Here's something more vague but also playful, per our style :)

jameslibscie: To me this still feels like a promise that we might not be able to make. Also, let's say we charge €5, I'm guessing there are plenty of countries where coffee still costs < €1.

How about, "We intend to offer small one-time fees for lifetime service."

chartgerink: Also difficult as we're still developing the business model. Maybe we should just delete the second sentence altogether then? Seems like a risky move then.

jameslibscie: Removed!

PR #209, opened 2020-08-05
Add protocol handler. Closes #176

### LINKED ISSUE #176: Open in Hypergraph
Any `hyper://` link should be opened in hypergraph.

- [ ] Register `hyper://` protocol handler
- [ ] Open given link in Spotlight search

juliangruber:
This makes it so `hyper://` urls open in Hypergraph :)

Try with the link [on this page](https://jsbin.com/sojuvokoxu/edit?html,output).

Closes #176.

## TODO

- [x] settle on exact protocol to use
- [x] doesn't yet work on mac


jameslibscie:
Awesome!

juliangruber:
> Awesome!

Did you test it or just like the change?

jameslibscie:
> > Awesome!
> 
> Did you test it or just like the change?

Tested it and works great :)

juliangruber:
ok. please submit a review next time please :)

juliangruber:
> 1. When I open the link of my own profile, I can select Hypergraph but it doesn't do anything? (does not jump to profile page for example).

I forgot to mention that you need to install the app for this to work. On macOS that means moving it to your Applications folder, and opening it once before clicking on the link. Could you try again?

juliangruber:
> 1. When I open the link of my own profile, I can select Hypergraph but it doesn't do anything? (does not jump to profile page for example).

The currently implemented behavior is for it to open the find modal, with the link prefilled. Otherwise we're missing a loading state for this..

juliangruber:
> 2\. If I try to "Find" my own profile in Hypergraph directly, it gives me the following error (not on release version)

This is unrelated to the PR I think, could you please open a separate issue?

juliangruber:
> 1. Would it be an idea to use a custom protocol? E.g., `p2pcommons://<key>` or `p2pcommons://<key>+<version>`? If we have `hyper://` I see us running in to issues with people who run Beaker as well, and opening all `hyper://` links in Hypergraph might create problems? I realise this is extra work as the browser doesn't allow custom protocols by default and requires us to [edit the registry](https://glebbahmutov.com/blog/electron-app-with-custom-protocol/).

We're editing the registry already for this to work. We could use a custom protocol yeah, but since the spec uses `hyper://` it would be confusing to use something else. What do you think?

chartgerink:
1. Same behavior as before --- Hypergraph doesn't do anything on my end (macOS 10.15.6). Also no messages in devtools.

2. Sure :)

3. True 😊 Could we do both? It seems like that would allow us to create stable links that would never open in Beaker as well, to use throughout the Vault emails and blog posts, but still allow people to open `hyper://` if they so choose.

juliangruber:
> 3\. True 😊 Could we do both? It seems like that would allow us to create stable links that would never open in Beaker as well, to use throughout the Vault emails and blog posts, but still allow people to open `hyper://` if they so choose.

Technically yes, but strategically I'm not sure. Why are we then not using the `p2pcommons://` protocol in the spec too?

juliangruber:
> 1. Same behavior as before --- Hypergraph doesn't do anything on my end (macOS 10.15.6). Also no messages in devtools.

Dang. I tested this on Windows and it works there, but not yet on macOS.

chartgerink:
> > 3. True 😊 Could we do both? It seems like that would allow us to create stable links that would never open in Beaker as well, to use throughout the Vault emails and blog posts, but still allow people to open `hyper://` if they so choose.
> 
> Technically yes, but strategically I'm not sure. Why are we then not using the `p2pcommons://` protocol in the spec too?

I was thinking along the lines that not all `hyper://` URLs have to be opened in eg Hypergraph, but all `p2pcommons://` would. As a result, if we say `hyper://` opens in Hypergraph and somebody (accidentally) sets it as the default handler it can create the situation where the app doesn't open something and the friction may be attributed to Hypergraph. Also, if somebody sets a different app as the default handler for `hyper://` (eg Beaker), the link wouldn't open in Hypergraph anymore.

But if we would use a separate protocol (eg `p2pcommons://` or more app-specific `hypergraph://`) we disambiguate these situations. If people wouldn't want to disambiguate where it opens (i.e., don't care), they could use `hyper://` and get all their suggested apps (or preferred default).

Technically the URL of the Hyperdrive is still `hyper://`, which is why it is included as such in the module spec. I do agree that if we choose to use `p2pcommons://` to allow opening modules in a p2pcommons-compliant app, we could add it to the interoperability spec. In that sense, we could also be generic and use `hypergraph://` for our specific scenario and leave it out of the specs entirely. 

I'm not set on it either way, I am thinking through how we might prevent confusing situations where people expect something to open in Hypergraph and it doesn't end up doing so for whatever reason.

What are your further thoughts on this?

jameslibscie:
I think you make a good point @chartgerink. In a way, the more successful we are, the more it could impede other hypercore applications from working correctly and vice versa. Perhaps this is something we should discuss with the hypercore team.

juliangruber:
That makes sense, thanks @chartgerink! I really like `hypergraph://`, because it makes immediate sense for the browser to ask whether it may open `hypergraph://` links with `Hypergraph`. My issue with `p2pcommons://` is that we haven't mentioned it in the app yet at all, except for in two dialogs.

juliangruber:
> I think you make a good point @chartgerink. In a way, the more successful we are, the more it could impede other hypercore applications from working correctly and vice versa. Perhaps this is something we should discuss with the hypercore team.

I think it's more that there are some hyper applications that can open anything, like Beaker, and some that only support a certain type, like Hypergraph.

jameslibscie:
> > I think you make a good point @chartgerink. In a way, the more successful we are, the more it could impede other hypercore applications from working correctly and vice versa. Perhaps this is something we should discuss with the hypercore team.
> 
> I think it's more that there are some hyper applications that can open anything, like Beaker, and some that only support a certain type, like Hypergraph.
> 
> Beaker eg would have no issue opening a p2pcommons module.

Yeah, but the association with Beaker would impact us negatively, because we'd prefer the link to open in Hypergraph. Conversely, an association with Hypergraph with impact Beaker negatively as non-p2pcommons hyper links wouldn't open in Beaker anymore.

I'd be for using `hypergraph://` protocol, but in a separate issue, as it also requires adjustments on the Share modals and probably the Find modal too?

The implication would also be that links shared from Hypergraph then also wouldn't open in Beaker anymore, but I think that's OK.

juliangruber:
SGTM! This is ready for final review @chartgerink, it works on mac now. @jameslibscie could you open an issue for the protocol change?

PR #214, opened 2020-08-05
Fix anchor styling

### LINKED ISSUE #212: Cancel button on Add Content screen incorrectly aligned
image

I broke this with my changes to `Anchor`. Want me to fix it @juliangruber?

jameslibscie:
Restores styling for all links except those in the welcome modals. Closes #212 

I tried two approaches that work:
1. Add `inParagraph` prop for `Anchors` that removes `display: inline-block`
2. Created styled `p` as `Paragraph` specifically for `welcome.js` that sets `display: inline` for all `a` within `Paragraph`

I think the second is cleaner, but the disadvantage is that this is then only applies to the welcome modals and can't be used without code duplication outside. See earlier commit for approach 1.

### DIFF
--- a/app/components/anchor.js
+++ b/app/components/anchor.js
@@ -8,6 +8,7 @@ const StyledAnchor = styled.a`
   color: ${props => (props.disabled ? gray : white)};
   border-bottom: 2px solid
     ${props => (props.disabled ? gray : props.color || purple)};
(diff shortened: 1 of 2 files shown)


juliangruber:
Closing in favor of https://github.com/hypergraph-xyz/desktop/pull/215

PR #218, opened 2020-08-07
Draft for onboarding tour

### LINKED ISSUE #91: integrate reactour
https://reactour.js.org/

https://www.notion.so/libscie/807068bc6784438d991660573ee48bd8?v=ddad829b5bf547e888301064053d90d3&p=a5b6118e83604825a66c89d5335a19db

jameslibscie:
A few issues that still need solving:

**Input is blocked while tour is open**
- [x] For some reason the Find URL and the Add description inputs are blocked while the tour is open, even if the element is focused on by the tour.

**Reopen tour menu item -> move from left menu to top menu**
- [x] Not sure how to add the menu item for reopening the tour as the state that stores whether the tour should be open or not is not available there. For dev purposes I've added it as a regular menu item. This obviously isn't the idea ;)

**Reopening to certain page + tour step upon backup (nice to have)**
- [ ] If the tour is open when backing up, the user should be taken back to the same page (we should probably do this even when the tour is off) and the tour should be restarted. Solving the menu item problem probably makes this easy.

**Other**
- [x] Text adjustments - see below
- [x] Tour jumps back to step 1 when clicking Add a description...
- [x] Feed -> Add child, Profile -> Add child, Following -> Unfollow and Drafts -> Open content steps don't work properly unless the user is hovering in the right place. Would be nice to trigger this hover.
- [x] Fix text overflow for URL
- [x] Check Profile Save & Share button behaviour
- [x] Text seems too big / spacious atm
- [x] Disable animations
- [x] No HTML in the text, even though we'd like to display links and maybe italicise certain text - otherwise apostrophes for actions and enable text selections
- [x] Bug: Reopen tour -> Find -> Remote profile (not yet followed) -> after clicking follow it focuses on the wrong element
- [ ] ❓ Explain network status indicator <- already think we have a ton of steps and you can already learn about this feature by hovering
- [ ] ❓ Justify align text <- doesn't look good
- [ ] ❓ Quit with any key on final step

See https://www.notion.so/Onboarding-tour-daff9ec3da664f4286dd12216cd6ed02 for text. ✔ = (more or less) implemented. There are some deviations from the text in Notion due to the way Reactour works.

Closes #91


jameslibscie:
@chartgerink I'll try to solve the page->modal->page problem before I go on holiday. If I succeed, perhaps then you could add the rest of the text if you have time next week? I think we'll need @juliangruber for the menu item.

chartgerink:
> @chartgerink I'll try to solve the page->modal->page problem before I go on holiday. If I succeed, perhaps then you could add the rest of the text if you have time next week? I think we'll need @juliangruber for the menu item.

I'm not really at the level to implement this I think 😅 (also looking at the code right now)

jameslibscie:
@chartgerink All the text is now implemented, but not everything works 100%. Will update the OP with a list of things that need doing when I'm back. Perhaps you could test it a bit and see if there are any more issues?

jameslibscie:
@juliangruber I'm guessing my changes to `package.json` and `package-lock.json` should be reverted, right?

chartgerink:
@jameslibscie thanks for implementing this up to here! Here's a list of issues I found, separated by tour (checkboxes for tracking). I checked off those that you already mentioned.

# General

- [ ] Animations between steps can be visually distracting and intense for people. Can we turn these off completely? (we do not maintain a `prefers-reduced-motion` trigger so better to reduce motion altogether for now)
- [ ] Can only exit with `ESC` or the exit button. Could we add that it quits with any key if on final step of a tour part?
- [ ] Reopen tour reopens after I also then open the screen I want a tour of. That isn't super clear of itself as the action "Reopen tour" implies it opens it up immediately. 
- [ ] We do not explain the network status indicator?

## Feed (without content/first open)

### Operations

- [ ] Step 2 - Cannot input text when I open the chat

### Text

- [ ] Step 1 - "Here's a tour for you" reads a bit awkward. Maybe "Welcome to the Hypergraph tour! You can close the tour and re-open it later through the Help menu"?
- [ ] Step 2 - Add a few different skin colours for diversity?
- [ ] Step 3 - "Let's get started..." reads somewhat lethargic. Could drop it and retain messaging
- [ ] Step 4 - Do we want to introduce `hyper://` URLs here? They seem to already be a big part of the find modal

## Feed (with content)

### Operations

- [ ] Step 2 - Oddly positioned and does not focus the add content button
Screenshot 2020-08-13 at 09 56 36

## Profile

Screen Recording 2020-08-13 at 09 39 10 2020-08-13 09_41_57

### Operations

- [ ] When I edit the description the tour jumps back to step 1 of the profile tour
- [ ] When I click save it skips to something it seems and then goes to next step?

### Text
- [ ] Actions are unclear in text (blend in) -> maybe add apostrophes? ('Back up database' in menu bar)

## Find modal

- [ ] Text overflow
Screenshot 2020-08-13 at 09 42 58
- [ ] Cannot select tour text
- [ ] Cannot input text while tour is open

## Following

### Operations

- [ ] Step 3 - Unfollow description in odd position (due to hover state only?)
Screenshot 2020-08-13 at 09 52 53

## Followed profile

### Operations

- [ ] Step 6 - Does not focus on follow button
Screenshot 2020-08-13 at 09 52 09

juliangruber:
> @juliangruber I'm guessing my changes to `package.json` and `package-lock.json` should be reverted, right?

Yes please, because the sdk-js version referenced is newer than the `0.7.0` release

juliangruber:
> @chartgerink I'll try to solve the page->modal->page problem before I go on holiday. If I succeed, perhaps then you could add the rest of the text if you have time next week? I think we'll need @juliangruber for the menu item.

I can take a look at this :+1:

jameslibscie:
@chartgerink Thanks for the detailed review! I've added your points to the OP. The Reopen tour page is just a temporary thing for testing until we have the menu bar item which will have the behaviour you expect.

chartgerink:
After discussing with @juliangruber, we decided to simplify the tour. It felt more like a tutorial and we boiled it down to some of the core bits now.

This was superbly easy to do thanks to the clear code @jameslibscie wrote. I hadn't worked with tour but was able to expand it in no time. I made a local copy of the tour notes for reuse at a later time.

chartgerink (approved):
Ready to merge.

Only note: Would need to add one step for #276

juliangruber:
@chartgerink do you want to take this over and create a new PR?

chartgerink:
Superceded by #283

PR #228, opened 2020-08-19
Add draft editing. Closes #69

### LINKED ISSUE #69: Update content module
**User story**
As a content creator
I want to update the title and description of my content
so that I can correct mistakes and make adjustments as work progresses

**Acceptance criteria**
- [ ] The content title is editable
- [ ] The description is editable
- [ ] It is possible to cancel in-progress changes
- [ ] It must not be possible for users to enter a content title that does not adhere to the specs (at the version currently implemented by the SDK version used)
- [ ] If the content does not have a description, the user is prompted to add one (for example, the text "Add a description..." is displayed greyed out with some kind of visual cue to indicate editability)

juliangruber:
Finally!

image

Closes #69


jameslibscie (changes requested):
Works great, except for a couple of small things:
- [ ] Add to profile button on edit page doesn't add to profile, but just saves as draft
- [x] Removing the parent doesn't do anything
- [x] Changing the parent adds it, so you then have 2 parents
- [ ] If I set the same parent again, I get the `parents_registered` error, which I think is triggered by the original version no longer being on the profile (I've edited the previous parent several times)

I think the last 3 are probably all fixed by the same change.

juliangruber:
> * If I set the same parent again, I get the `parents_registered` error, which I think is triggered by the original version no longer being on the profile (I've edited the previous parent several times)

I can't reproduce this, maybe this has been fixed already by the previous commit

jameslibscie (changes requested):
I'm wondering whether this is an existing bug or not, but I've been able to get modules to display twice - both on Profile and in Drafts, just by doing the following:
1. Create module and add to profile (either directly or via content page)
2. Restart Hypergraph

No editing necessary...

juliangruber:
> I'm wondering whether this is an existing bug or not, but I've been able to get modules to display twice - both on Profile and in Drafts, just by doing the following:
> 
> 1. Create module and add to profile (either directly or via content page)
> 2. Restart Hypergraph
> 
> No editing necessary...

I can reproduce this in `main` too, but only sometimes. Could you open another issue please?

jameslibscie:
> > I'm wondering whether this is an existing bug or not, but I've been able to get modules to display twice - both on Profile and in Drafts, just by doing the following:
> > 
> > 1. Create module and add to profile (either directly or via content page)
> > 2. Restart Hypergraph
> > 
> > No editing necessary...
> 
> I can reproduce this in `main` too, but only sometimes. Could you open another issue please?

Will do! Then I can approve this one

juliangruber:
> I can reproduce this in `main` too, but only sometimes. Could you open another issue please?

This is a SDK bug. Using some `console.log`s, I noticed that the same module is returned with different versions in two consecutive `p2p.listContents()` calls:

First call, where it's listed as draft (`version = 11`)

image

Second call, where it isn't (`version = 4`)

image

jameslibscie (dismissed):
Will create an SDK issue for the draft/registered split

dpaez:
I think that somehow we entered wrong/bad data to the localdb. We can add some checks on `listContent` method to filter values and return the one with the latest modified time or latest version. Which one? It should be the same.

On the other hand, https://github.com/p2pcommons/sdk-js/pull/221 is an attempt to prevent this.

PR #233, opened 2020-08-21
Various text improvements following from user testing

### LINKED ISSUE #202: Feed text improvement suggestion
Now, Feed opens first instead of Drafts, which means that the instructions for using the global add content button are now harder to find. In addition, the text "Share content" is not correct since sharing content doesn't make it appear on the feed - adding it to the profile does.

If user has no content at all:
> Add content to your profile using <hex + icon> or follow someone to fill up your feed 🙌

If user has draft content:
> Add content to your profile or follow someone to fill up your feed 🙌

jameslibscie:
Changes:
- Consistent use of "step" instead of "part" in welcome modals
- "Finding a journal" -> "becoming accepted by a journal"
- "Hypergraph Vault is our cloud service"
- "Nudges" on all pages with icon hints on how to get started
- Removed "No main" text on edit content form to make it visually clearer that nothing is selected

@psobrakseaton there are some small design implications because I'm including the Find icon on other pages for the first time. It's probably enough to just check these screenshots:

image
image
image

Curious for opinions! If there are any major doubts, I will just scrap these changes as this is technically non-planned work that is only worth the effort if we can merge fast imo.

Closes #202


app/components/content/edit-form.js
```diff
@@ -257,7 +257,7 @@ const EditForm = ({
           value={main}
           onChange={ev => setMain(ev.target.value)}
         >
-          <option value=''>No main</option>
+          <option value='' />
```
juliangruber, line 260: It looks to be common UX practice to have a word for the blank option: https://ux.stackexchange.com/questions/36633/best-wording-of-blank-option-for-an-optional-select. Why did you change this? Maybe we could find a better replacement.

app/components/feed/feed.js
```diff
@@ -55,7 +74,7 @@ export default ({ p2p }) => {
               <>
                 {contents.length
                   ? 'You’ve reached the end! ✌️'
-                  : 'Share content or follow someone to fill up your feed 🙌'}
+                  : <>Add content with <StyledAddContent /> or <StyledSearch /> Find someone to follow 🙌</>}
```
juliangruber, line 77: To me the mixture of multiple icons and emojis is visual overload. What about either removing the emojis or scaling down the icons so they don't stick out of the text flow?

app/components/feed/feed.js
```diff
@@ -5,6 +5,25 @@ import { encode } from 'dat-encoding'
...
+import styled from 'styled-components'
+import AddContent from '../icons/add-content.svg'
+import Search from '../icons/search-icon-1rem.svg'
+
+const StyledAddContent = styled(AddContent)`
+  vertical-align: middle;
+  transform: scale(0.75);
+  position: relative;
+  top: -2px;
+`
```
juliangruber, line 17: This can by DRYed up. Let's revisit it once we decided on a strategy from a visual perspective.

app/components/welcome/welcome.js
```diff
@@ -82,8 +82,8 @@ const dialogs = [
       <Heading>As-you-go, not after-the-fact</Heading>
       <p>
         You are probably used to doing research, writing a full paper, and then
-        going through the laborious process of finding a journal. With
-        Hypergraph, you can publish each step of your research as-you-go.
+        going through the laborious process of becoming accepted by a journal.
+        With Hypergraph, you can publish each step of your research as-you-go.
```
juliangruber, line 86: :+1:

app/components/welcome/welcome.js
```diff
@@ -105,7 +105,7 @@ const dialogs = [
       </Illustration>
       <Heading>How Hypergraph works</Heading>
       <p>
-        Each part of your research is its own publication, whether it's a
+        Each step of your research is its own publication, whether it's a
```
juliangruber, line 108: :+1:

app/components/welcome/welcome.js
```diff
@@ -174,9 +174,9 @@ const dialogs = [
           between.
         </p>
         <p>
-          Hypergraph Vault makes your work available even when your computer is
-          offline. Plus, we're working with libraries on archiving it for future
-          generations! 👵🏾👨🏻👶
+          Hypergraph Vault is our cloud service that makes your work available
```
juliangruber, line 177:
```suggestion
          Hypergraph Vault is our service that makes your work available
```

Cloud has a negative connotation, especially in the P2P space. I'm also not sure this level of technicality is helpful here. Any reasons for specifically mentioning it?

Also, we aren't sure it even is going to be a cloud service, at all. It can run just fine on Chris' laptop, and maybe it will end up in some data centers and never the cloud.

juliangruber (changes requested):

app/components/content/edit-form.js
jameslibscie: There were two reasons:
1. Consistency -> follows from dropdown is also empty when there is no option selected
2. Users sometimes seem to skip over this dropdown, I was thinking perhaps because there is already something filled in.

Given your link, I'd prefer to then change both to "None" rather than leave them empty. This is common and apparently also preferred UX-wise.

app/components/welcome/welcome.js
jameslibscie: I didn't know about the negative connotation. How about "hosting" or "storage" then? I'm looking for simple ways to illustrate the difference between using the vault and not using the vault without literally explaining everything in detail.

For example, during one of last week's tests, it was thought or considered for a while that the vault would also be p2p-based and might require users to offer bandwidth and storage. The word "cloud" came up as something that is nowadays understood as something like "we'll take of your files for you" and implies safety.

app/components/feed/feed.js
jameslibscie: I kinda missed the emoji, so I've tried smaller icons. What do you think?

app/components/feed/feed.js
chartgerink: Could you explain the acronym? Tried googling but to no avail.

jameslibscie: "don't repeat yourself"

app/components/profile/following.js
```diff
@@ -100,7 +110,7 @@ const Following = ({ p2p }) => {
             title={
               following.length
                 ? 'You’ve reached the end! ✌️'
-                : 'You’re not following anybody yet 😅'
+                : <>You’re not following anybody yet. 😅 Use <StyledSearch /> Find to look up a profile</>
```
chartgerink, line 113:
```suggestion
                : <>You’re not following anybody yet. 😅 Use "<StyledSearch /> Find" to look up a profile</>
```

I was confused by this as I read "Find to look up a profile" 😅

app/components/welcome/welcome.js
```diff
@@ -82,8 +82,8 @@ const dialogs = [
       <Heading>As-you-go, not after-the-fact</Heading>
       <p>
         You are probably used to doing research, writing a full paper, and then
-        going through the laborious process of finding a journal. With
-        Hypergraph, you can publish each step of your research as-you-go.
+        going through the laborious process of becoming accepted by a journal.
```
chartgerink, line 85:
```suggestion
        going through the laborious process of getting accepted by a journal.
```

Becoming accepted sounds very existential, as researcher I'd use this phrasing (publication gets accepted).

app/components/welcome/welcome.js
```diff
@@ -169,18 +169,17 @@ const dialogs = [
         </Illustration>
         <Heading>Introducing the Vault</Heading>
         <p>
-          We use peer-to-peer technology: Hypergraph is powered by the people
-          using it! Share your content as you'd like, without anybody getting in
-          between.
+          We do not want to be gatekeepers. We chose peer-to-peer technology so you can
```
chartgerink, line 172: YAAAS!

chartgerink (approved):
Minor comments that you can do with as you see fit :) Nice work!

app/components/drafts/drafts.js
```diff
@@ -11,9 +11,11 @@ import sort from '../../lib/sort'
 
 const StyledAddContent = styled(AddContent)`
   vertical-align: middle;
-  transform: scale(0.75);
+  transform: scale(0.6);
   position: relative;
   top: -2px;
+  margin-left: -6px;
+  margin-right: -6px;
```
juliangruber, line 18: you could also remove the white space from the text surrounding it instead:

image

app/components/feed/feed.js
juliangruber: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself. It's more of a programming joke / slang I'd say to use it this way

app/components/feed/feed.js
juliangruber: image

This doesn't read well to me, I'd like to hear @psobrakseaton's perspective.

1) The add content icon is to the right while the find icon is to the left
2) What about removing the "with" from "add content with"?

app/components/profile/following.js
juliangruber, line 113:
```suggestion
                : <>You’re not following anybody yet 😅 Use <StyledSearch /> Find to look up a profile</>
```

Didn't we say in the welcome modals that emojis replace punctuation?

juliangruber: I noticed this too. Not sure "Find 🔎" is enough. @psobrakseaton what do you think?

app/components/profile/profile.js
```diff
@@ -348,7 +358,7 @@ const Profile = ({ p2p }) => {
             title={
               contents.length
                 ? 'You’ve reached the end! ✌️'
-                : 'No content yet... 🤔'
+                : <>No content yet... 🤔 Use <StyledAddContent /> to add something!</>
```
juliangruber, line 361: image

To my eyes there's just too many icons in here

juliangruber (changes requested):

app/components/profile/profile.js
psobrakseaton: I also think that we need another word, like "Use 🔎 Find to input a URL". 

I think the emojis are cute, but we should try to limit them if we are already using icons. It is throwing off my perception of the punctuation and making it hard to parse the sentence.

psobrakseaton: I think we either need to separate something by lines, or go for a simpler solution only using the add-content button.

psobrakseaton:
@jameslibscie can you post a list of all the changed messages in a single comment?

psobrakseaton:
We're solving this during a phone call...

juliangruber (approved):

PR #244, opened 2020-08-26
Set directory upon register / deregister

### LINKED ISSUE #234: Open folder and open file behaviour inconsistent when registering/deregistering
## Description

For registered content, the Open folder and file buttons point to the versioned copies. For unregistered content, these point to the unversioned folder.

However, if I am on unregistered content and add it to my profile, these buttons continue to point to the unversioned folder. Likewise, if I am on registered content and remove it from my profile, these buttons continue to point to the versioned folder.

Especially when the read-only files are introduced, this could become very confusing as it would allow you to edit registered content (but this change would later become invisible again). Similarly it could prevent you from editing unregistered content.

## To Reproduce

1. Create content with at least a main file and save as draft
2. Click open folder and open the file -> see that it points to unversioned
3. Add to profile
4. Click open folder and open the file -> see that it points to unversioned
5. Open profile
6. Re-open content
7. Click open folder and open the file -> see that it points to versioned
8. Remove from profile
9. Click open folder and open the file -> see that it points to versioned

## Expected behavior

When registering and deregistering content, the folder and file links should update to match the state of that content.

## Actual behavior

The links don't update until the page is closed and revisited.

## System

- OS (incl. version): Windows 10
- Hypergraph Desktop version: 0.6.0

jameslibscie:
Updates the directory after registering to include the version; after deregistering to exclude the version. This might not suffice in multiple author situations if you're a contributing author, but not the owner.

Ideally, calling register / deregister would update the global version on that page, triggering a directory update. This seemed like a rather big update. If you prefer that solution @juliangruber, feel free to close this :)

Closes #234


app/components/content/content.js
```diff
@@ -303,6 +303,8 @@ const Content = ({ p2p, contentKey: key, version, renderRow }) => {
                   )
                 } finally {
                   setIsUpdatingRegistration(false)
+                  const directory = await getContentDirectory({ key: encode(content.rawJSON.url), version: content.metadata.version })
```
juliangruber, line 306: This doesn't really belong in the `finally` as it should only be executed if the (de)registration succeeds. Should we move this up?

juliangruber: Did you run `npm test` on this?

juliangruber (changes requested):

juliangruber (approved):

PR #247, opened 2020-08-28
Always register latest version

jameslibscie:
Uses in-progress fix https://github.com/p2pcommons/sdk-js/pull/229 to always `get()` latest version and register the latest version.

Since the content page doesn't automatically switch to the registered version (see #234), leave the content page and re-open via the profile to verify that indeed the updates to files have come through. Or check the profile's metadata :)


app/components/content/content.js
```diff
@@ -295,16 +295,18 @@ const Content = ({ p2p, contentKey: key, version, renderRow }) => {
               onClick={async () => {
                 setIsUpdatingRegistration(true)
                 try {
+                  await p2p.sync(content.rawJSON.url, directory)
```
juliangruber, line 298: There's a risk here: If the module updates after the page has been loaded, and you then hit `register`, you're _not_ going to register what you see on the screen. This might not be a problem, but it's still better to implement the watching/events for this than to sync at the latest moment.

juliangruber (changes requested):

app/components/content/content.js
jameslibscie: I think that's a completely acceptable risk atm. The only changes I expect to occur outside of Hypergraph now are file changes and users would expect those updates to be registered. Especially as if the user opens one of the files, they *would* see the latest version, so if you were to register a previous version, they'd be registering something *different* than what's on the screen - and could feel like they've lost data this way.

I completely agree that watching/events is better, as that would deliver the smoothest user experience, but in practice I expect that the difference will be very small.

chartgerink:
Is this something we still want to take up @juliangruber before release, later, or can we close this?

Seems like `main` is implementing this already?

juliangruber:
Good catch, this can be closed

PR #258, opened 2020-09-07
Delete versioned module

### LINKED ISSUE #219: Delete removes unversioned folder only
## Description

When I delete content, it removes the unversioned folder only.

## To Reproduce

1. Create content module with main
2. Add to profile
3. Note the hyperdrive key
4. Go to ~/.p2pcommons
5. Delete registered content
6. Check ~/.p2pcommons

## Expected behavior

I expect the application to delete the relevant folders on my computer.

## Actual behavior

It only deletes unversioned folders and ignores versioned folders.

## System

- OS (incl. version): macOS 10.15.6
- Hypergraph Desktop version: 0.5.4

## Additional context

I think this is legacy of the delete content button for the drafts, where it is always unversioned. When it's registered and entered through either the feed or the profile, it is a versioned folder. I think the issue is in [this line of code](https://github.com/hypergraph-xyz/desktop/blob/c56071841a52ade0ef5f424bf7522783c8b584ba/app/components/content/content.js#L335) and that it does not differentiate for registered and unregistered content.

chartgerink:
In an attempt to fix #219 I changed the code a tiny bit to specify the versioned module for removal, not the unversioned. Note: This does not remove the unversioned module at this point.

If merged, closes #219. Please feel free to close this PR if this is inconsistent with other development pipelines.


juliangruber:
Thanks Chris, but the delete signature has changed and only takes a key now: https://github.com/p2pcommons/sdk-js#delete

chartgerink:
Hadn't noticed the signature change for `p2p.delete()`, thanks for mentioning. 

I now no longer know how to fix and delete the versioned and unversioned ones :sweat_smile: (it doesn't based on the latest sdk version from main)

PR #263, opened 2020-09-10
Add author name(s) to follows from list. Closes #223

### LINKED ISSUE #223: Author missing from "Follows from" list
image

The module "test" is from Chris, the others I have locally. It would be great to be able to differentiate between the different authors. Maybe a format like "Name (Author)" or "Author: Name" could work.

juliangruber:
Closes #223.

image


psobrakseaton (approved):
Works well and solves a problem without additional design work! I like it!

chartgerink (approved):
Looks good! ✅ 

I think I discovered an sdk bug, because I unfollowed somebody but still saw their content in the follows from list? Approving because I assume it's `sdk` so I'll open an issue there if that's correct.

PR #269, opened 2020-09-15
Fix main file validations. Closes #249

### LINKED ISSUE #249: Handle disallowed main file extensions
The SDK now only allows certain open file extensions for the main file. If the user has a disallowed file type as the main file, the error is not yet handled by Hypergraph.

EDIT: The user should be encouraged to use the chat if they think something is incorrectly being disallowed.

juliangruber:
This prevents users from selecting main files that wouldn't pass the main file validations.

Closes #249.


chartgerink:
Nice catch :) 

In testing the implementation, I became a bit confused. I opened a .xls file, which got added to the list while it isn't a supported file format. There was no visual indication to teach me what I could add or why I could not select it as main file. I only understood this interaction because of previous knowledge about this.

Would it be possible to have some visual/text indication to say that these are not supported (not open) file formats? At the moment, there is little visual indication what's erroring and this would help. @psobrakseaton I discussed with @juliangruber and we'd appreciate your input on this.

It might also help to update the copy on top to include something along the lines of `Only supports open file formats` and link out to Wikipedia maybe?

Attached a screenshot.

Screenshot 2020-09-15 at 16 39 49

psobrakseaton:
> Nice :) Works on the technical level as it should, including text files that are of unknown extension 🙌
> 
> In testing the implementation, I became a bit confused. I opened a .xls file, which got added to the list while it isn't a supported file format. There was no visual indication to teach me what I could add or why I could not select it as main file. I only understood this interaction because of previous knowledge about this.
> 
> Would it be possible to have some visual/text indication to say that these are not supported (not open) file formats? At the moment, there is little visual indication what's erroring and this would help. @psobrakseaton I discussed with @juliangruber and we'd appreciate your input on this.
> 
> It might also help to update the copy on top to include something along the lines of `Only supports open file formats` and link out to Wikipedia maybe?
> 
> Attached a screenshot.
> 
> Screenshot 2020-09-15 at 16 39 49

I think having this message is a really simple, but incomplete solution to the problem. Having visual form errors and messages will be very important for accessibility in the future. 

I think having the link on the top is a good enough solution for now, but our whole approach to the forms needs to change in a future revision.

chartgerink:
@psobrakseaton I tagged you in the hopes that you could provide some ideas / designs on how to do this here. Could you help us move this forward? That is, how can we improve that here in this version of the form?

juliangruber:
> I think having the link on the top is a good enough solution for now

What do you mean by this?

chartgerink:
I think @psobrakseaton is referring to what I said:

> It might also help to update the copy on top to include something along the lines of "Only supports open file formats" and link out to Wikipedia maybe?

psobrakseaton:
> @psobrakseaton I tagged you in the hopes that you could provide some ideas / designs on how to do this here. Could you help us move this forward? That is, how can we improve that here in this version of the form?

I think with this task our options are quite limited. When I was testing this with proprietary file formats the error only occurs once we try to submit everything. Is there a way for the validation to happen before submission?

We basically can provide a list of supported file formats (the link to the compiled list), or not allow them to upload unsupported formats at the start and provide some sort of message that doesn't look like an error, but currently we don't have error messages designed beyond using system alerts.

juliangruber:
New plan: show a dialog when invalid files were tried to be uploaded, and list the files that couldn't be added

juliangruber:
What about this?

image

psobrakseaton:
> What about this?
> 
> image

I think that this is a good step forward with listing the items that will not be uploaded in the batch. That gives them time to review the files that are uploaded.

My one suggestion for now would be to remove the path and just have the file name.

chartgerink (dismissed):
LGTM! I like that we reiterate the need for open file formats to the user in the alert :)

PR #272, opened 2020-09-15
Manually trackPageView to get all URL changes

juliangruber:
This should list all navigation changes in Matomo, while before only `dist/index.html` was shown.

I wasn't able to test this yet, because the Matomo dashboard didn't show any new data.


juliangruber:
image

Yus it worked!

chartgerink (dismissed):
👍 Nice work @juliangruber!

I double checked and there's no private information leaking in the statistics either (e.g., own profile on different URL that can be used to identify user).

EDIT: Still valid with the `/profile` route that solves #273 :)

PR #276, opened 2020-09-17
Add discover page. Closes #274

### LINKED ISSUE #274: Discover screen
**Is your feature request related to a problem? Please describe.**
At the moment, users cannot see any example content and learn from other users on what Hypergraph is capable of, when they start out and have nobody to follow (presumably).

In SSB and Rotonde, it wasn't uncommon to lose people's interest because there was no example content. The cold-start issue might be one we also face with Hypergraph.

**Describe the solution you'd like**
The Vault has a feed of new content, that could be the basis of seeing content and find people to follow. If we'd have X most recent additions in a separate screen (e.g., "Discover") on its own, people could find content and people without clogging their feed directly. 

We could for example only take the ten most recent, and download the next ten if people hit the bottom of the screen to create an infinite scroll (if we want to).

We could limit it to content only so no design for profile cards would be needed in the feed (or we could use the one from "Following", maybe?).

**Describe alternatives you've considered**
Adding some random profiles to the "Find" modal to click on. Might require more design work.

**Additional context**
This is a stretch goal for cycle ~6~ 5. During the walkthroughs, it appeared that people would like this.

juliangruber:
This adds a basic Discover page. Closes #274.

image

## TODO

- [x] It takes a bit for the modules to download. And some don't download at all. What is a sensible timeout? I just picked 5 seconds for now. In order for the modules to be clickable, their content files need to be downloaded, which prolongs the clone time.
- [x] Only one of the latest 10 content modules successfully downloads currently. Shall we add more example content? Another option is to keep cloning until 10 modules have been downloaded, which could however take even longer.
- [x] When you click on a module, the back arrow works, however the "Discover" menu bar entry isn't selected any more. What shall we do about that?

image

- [ ] There needs to be an empty state for when you don't have connectivity.


chartgerink:
😲 this looks superb.

> It takes a bit for the modules to download. And some don't download at all. What is a sensible timeout? I just picked 5 seconds for now. In order for the modules to be clickable, their content files need to be downloaded, which prolongs the clone time.

Five seconds seems okay for now. Could we continue downloading async, so that when people revisit the screen it shows more? Or could we start the download upon start, so the discover screen is ready upon click?

> Only one of the latest 10 content modules successfully downloads currently. Shall we add more example content? Another option is to keep cloning until 10 modules have been downloaded, which could however take even longer.

We definitely will have to start adding more example content. Any idea why the replication works for 1/10? Might be part of  the issue we ran into in our debugging session as well?

>When you click on a module, the back arrow works, however the "Discover" menu bar entry isn't selected any more. What shall we do about that?

Standard process is for it to be highlighted right? Is there a challenge to do this here?

> There needs to be an empty state for when you don't have connectivity.

Ah yes! Could we alert them with a text that says `Please connect to the Internet` or something? When they press okay they are reverted to another screen, e.g., Feed.

juliangruber:
> Could we continue downloading async, so that when people revisit the screen it shows more?

We can try that, yes!

> Or could we start the download upon start, so the discover screen is ready upon click?

That's also an idea.

I think however that both are in-between solutions, and letting users select what to download, and when, and letting them see how much has been downloaded is better.

> Any idea why the replication works for 1/10? Might be part of the issue we ran into in our debugging session as well?

Afaik only 1 out of the latest 10 modules showed as 100% in the seeder dashboard.

> Standard process is for it to be highlighted right? Is there a challenge to do this here?

The challenge is that now the seeder page would also need to have its own "look at an individual content module page", although it doesn't add any value.

> Could we alert them with a text that says `Please connect to the Internet` or something? When they press okay they are reverted to another screen, e.g., Feed.

Ah, so you mean to open a dialog when the Discover page is opened without internet? Or what about greying out that element and showing a tooltip saying why it's not available?

chartgerink:
> I think however that both are in-between solutions, and letting users select what to download, and when, and letting them see how much has been downloaded is better.

Agreed! Noted in the potential work items as "Download progress" and "Sparse download" 

> The challenge is that now the seeder page would also need to have its own "look at an individual content module page", although it doesn't add any value.

Then let's leave it as is. Can always adjust later if people find it confusing.

> Ah, so you mean to open a dialog when the Discover page is opened without internet? Or what about greying out that element and showing a tooltip saying why it's not available?

Greying out sounds like a straightforward way 👍

psobrakseaton (dismissed):

juliangruber:
> Greying out sounds like a straightforward way 👍

@psobrakseaton you just approvied this PR, but did you see this?

psobrakseaton:
> @psobrakseaton you just approvied this PR, but did you see this?

Totally didn't see this...I was just testing the version that was up here. Can I retract my approval?

psobrakseaton:
Thanks!

juliangruber:
@psobrakseaton I have a question about loading states:

Currently, when you look at the discovery feed, the module files _won't_ be downloaded, as that most of the time takes too long, to work for a whole list. Now when you then click on a module, we need to download the files, otherwise the file list on the content doesn't work.

How should this loading state then look like? Show a loading indicator over the whole screen? Show one below the header rows?

juliangruber:
The most useful would be if the content page was fully accessible, except for the file operations (eg listing files, maybe open folder)

psobrakseaton:
This is a really good point.

I agree that everything except the downloading files should be accessible...I will get to work on some designs...

psobrakseaton:
Ok, so here are two directions we can go...Once we decide on one I will give you the guidelines for implementation and make sure you have the assets.

I think my preferred method is this: 
spinners

I like that with the spinners you can see potentially how many files are loading and it will make you more willing to wait for the results. I could imagine that perhaps this isn't as easy to implement.

So there is also this with the dots over the files areas:
dots

juliangruber:
> I like that with the spinners you can see potentially how many files are loading and it will make you more willing to wait for the results. I could imagine that perhaps this isn't as easy to implement.

Right now unfortunately we have no way of knowing how many files there will be

juliangruber:
It's more tricky with the main file: We know its name, but it won't be immediately available. So there's this in between state where it's not loaded yet, but at least we know the name. Whether it's useful to expose the name in this case is uncertain.

psobrakseaton:
Ok, I here's my thinking, we know that if someone is seeing the content at all then there is a main file, but that doesn't necessarily mean that there are supporting files. I think we should do basically what I had above, but show the main file loading like this:

Screen Shot 2020-09-28 at 17 32 18

Then, if we could load atomically like we do here as we get more information that would be the best I think:

Screen Shot 2020-09-28 at 17 34 10

A sort of "rolling" loading cycle if that make sense...

juliangruber:
I think we're still not quite there yet with the module loading state:

1) If we decide not to download module files right away on screens like the discovery screen, the user will not be able to tell. If they look at the list of modules, they have no indication of which module has been downloaded, not yet. or is downloading. Right now, before this change, every module that you see in a list is downloaded and available, which is very clear.

2) @psobrakseaton we don't get any information about which files are loading. We know that either
  a) The metadata is there but the files are still loading
  b) The metadata is there and the files have completely been downloaded
We don't know any in between.

psobrakseaton:
Text for the empty feed view for the Discover tab ---> Only available when connected to the Internet ☝️ 

Reference image:
Screen Shot 2020-09-29 at 14 03 00

psobrakseaton (approved):
Works perfectly online and offline!

chartgerink (approved):
My comments were resolved during office hour 🚪 

I also tested the offline display and it's nice 😊

PR #285, opened 2020-09-23
Fix author loading or unavailable. Closes #266

### LINKED ISSUE #266: Handle case where author isn't available on the network
Eg you look at a module with multiple parents and one of the parents isn't available (locally or over the network).

juliangruber:
This shows a loading indicator for up to 5s, until "Author Unavailable" will be shown.

Closes #266.

You can test this by following author A who was registered content A with author B who isn't available on the network. Or you can look at these screenshots:

### Content list

image

image

### Content screen

image

image


psobrakseaton:
Yay! Happy to see this implemented! I have two requests based off the screenshots.

1. I will make a new loading indicator for the content list, I thought it could work, but it is too small.
2. I think I want the version in the content to be pushed down by like 4px, I'll wait until the build comes out and then I can give exact figures.

psobrakseaton:
Ok, so I ran some tests in Figma and I think you are right about having the space before the spinner so that it takes up the same visual space as the name will.

I modified the SVGs to make them essentially like a single font character with the same "line-height" for the space they will occupy and the only padding amounting to kerning. These can be placed in the string with or without spaces at our discretion.

For now lets have the space in front.

Here is a visual example of how there is no spacing between the strings and the SVG:

content-block-example
content-page-example

I'll send you the new SVGs via Mattermost!

juliangruber:
> I notice that from the code there is a 50% chance for Hypergraph to attempt cloning the profile. In my tests this also occurs if I already have the profile, which results in a random load animation for information already on the device? It also keeps loading ad infinitum in that case.

This was just for patch to debug it 😅 The commit before that doesn't have that logic

psobrakseaton:
That is looking pretty freaking awesome already! I think with the new SVGs it is gonna be even better! Very stoked on this...

Also, the colored avatars really are awesome...

chartgerink (approved):
Checked with 06cc5726343048f7606966c1d45552eb11d6b6a3 and works super nice :)

juliangruber:
The new SVGs have been added. You can try the 50% debug version in https://github.com/hypergraph-xyz/desktop/pull/285/commits/2efc3c80eeac71ea891978438439b8150fe780eb and the finished implementation in https://github.com/hypergraph-xyz/desktop/pull/285/commits/750d84d0e4057fc9798b82f813d1c7c84b1be2a1.

chartgerink (approved):
Really digging the atomic loading approach! Makes it feel very responsive.

psobrakseaton (approved):
Perfect! I'm glad we brought this back into the garage...I think the size is just about perfect on these now!

Thanks, Julian! <3

PR #288, opened 2020-09-24
Add/hover plus

chartgerink:
One of the walkthroughs mentioned that the inconsistency between the hover plus and the omnipresent add content button was bugging them. I talked with @psobrakseaton and he mentioned it was just legacy and indeed inconsistent.

I made an adjustment to have them be similar in style. I had to shrink it down a tad, because the original plus was 48x48px, and the omnipresent is 56x64px, which led to an overflow in the hover box.

# Screenshot of implementation

Screenshot 2020-09-24 at 11 57 24

## Overflow issue mentioned (not implemented)

Screenshot 2020-09-24 at 11 40 33


juliangruber:
Are there designs for how this is supposed to look?

juliangruber:
I'm not sure what the PR description means. Are you satisfied with how it is?

juliangruber (approved):
LGTM! Nice change

chartgerink:
> Are there designs for how this is supposed to look?

This was more emergent than designed. Coordinated with @psobrakseaton and he approved too :) 

> I'm not sure what the PR description means. Are you satisfied with how it is?

I like it better with the hex 😊

juliangruber:
Ok yeah makes sense to me too 👏

PR #289, opened 2020-09-24
Update: catch clone errors 

dpaez:
This PR is a little tweak that looks to populate the feed with some content and finish earlier. 

sdk `clone` method is prone to throw errors when content (`index.json`) is unreachable. This can be the case for specific version that are not shared at the moment, for example.

### DIFF
--- a/app/components/feed/feed.js
+++ b/app/components/feed/feed.js
@@ -24,12 +24,18 @@ export default ({ p2p }) => {
       const contentUrls = [
         ...new Set(profiles.map(profile => profile.rawJSON.contents).flat())
       ]
-      const contents = await Promise.all(
-        contentUrls.map(url => {
+
+      const contents = []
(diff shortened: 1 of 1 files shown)


app/components/feed/feed.js
```diff
@@ -24,12 +24,18 @@ export default ({ p2p }) => {
...
+
+      const contents = []
+      for (const url of contentUrls) {
+        try {
           const [key, version] = url.split('+')
-          return p2p.clone(encode(key), version)
-        })
-      )
+          const c = await p2p.clone(encode(key), version)
+          contents.push(c)
```
juliangruber, line 33:
```suggestion
          const content = await p2p.clone(encode(key), version)
          contents.push(content)
```

juliangruber (changes requested):
Just a tiny change, otherwise looks great! Thanks for submitting

PR #292, opened 2020-09-28
Fix credits position to always stick to bottom. Closes #194

### LINKED ISSUE #194: Footer doesn't stick to bottom
image

It should be below all content, but if there is more space, stick to the bottom of the window.

juliangruber:
Closes #194.

image


chartgerink (approved):

psobrakseaton (approved):
Works perfectly! Also got to do the tour, which was nice!

PR #293, opened 2020-09-28
Update copy and add link for supporting memberships

chartgerink:
I noticed the supporting membership didn't link out yet (logically), so I added the preview link to the blog post that is upcoming still. Once this post has been published, it will auto-resolve to the final link.

I also made a minor change to the copy to be more direct.﻿


app/components/welcome/welcome.js
```diff
@@ -274,12 +274,11 @@ const dialogs = [
...
-          at{' '}
-          <Anchor href='mailto:community@libscie.org'>
-            community@libscie.org
-          </Anchor>
-          .
+          . Want to financially support us? Become a{' '}
+          <Anchor href='https://blog.libscie.org/p/6569c48c-7f53-4c63-ab68-e654a54abe96/'>
+            supporting member
+          </Anchor>{' '}
+          for €70 a year.
```
juliangruber, line 281:
```suggestion
          </Anchor>.
```

The price might update more often than this up

juliangruber (changes requested):

app/components/welcome/welcome.js
chartgerink: Made a note in Notion so we don't forget to change this if the price changes.

PR #303, opened 2020-09-30
Use new SDK events for cloning

juliangruber:
Integrating https://github.com/p2pcommons/sdk-js/pull/252


chartgerink:
Thanks @juliangruber for this straightforward and logical implementation!

After testing this intensively tonight, it seems like the new `cloneOne` method is introducing issues that affect the experience in Hypergraph. I wouldn't merge this for the beta release as a result.

Specific issues:
* The find modal frequently stalls while loading a profile, plus the exit button for the modal doesn't work anymore (requires a hard reload to get out)
* The discovery page times out often, making it practically unusable.

Without these new methods, the resume integration still works (albeit silently, [see PR #252](https://github.com/p2pcommons/sdk-js/pull/252) for details). @dpaez and I tried a substantial amount of variations during testing to see whether we could try to resolve these issues, but the most straightforward fix remained using the `main` branch. The stalling probably happened because the content wasn't returned due to not getting a completion emitter (it kept emitting download progress for only an `index.json`).

We now know we resume the module download, even if we still do not show when a module is incomplete on disk. I spoke with @dpaez about ways to tackle this, and it seems like we can hotfix this when we add features, but couldn't tackle it here within the timeframe.

chartgerink (approved):

PR #304, opened 2020-10-01
Update tour copy and terms

chartgerink:
Some minor corrections to the terms upon re-reading, plus some additional text for the first screen on the welcome tour, to repeat that this is a beta version and people should be patient.﻿


app/components/tour/tour.js
```diff
@@ -25,9 +25,15 @@ const Tour = ({ onClose, ...props }) => {
         {
           content: (
             <div>
+              <p>Welcome to Hypergraph (Beta) 🥳</p>
               <p>
-                Welcome to Hypergraph (Beta) 🥳 We'll get you started with a
-                quick tour of the most important things.
+                The beta has some rough edges, so please be patient and kind to
+                it (especially when downloading files). Maybe take a break while
+                you're waiting.
```
juliangruber, line 32:
```suggestion
                it (especially when downloading files).
```

We don't have a problem with it downloading slowly

app/components/welcome/terms-of-use.js
```diff
@@ -327,9 +327,10 @@ export default () => (
           10365 Berlin (DEU)
         </p>
         <p id='40cb04cf-561d-4b56-9af8-67f7996dde81' className=''>
-          If we assess a takedown claim as valid (or are court ordered), we will
-          check our servers for copies of the relevant content and add it to our
-          block-list for any services related to Hypergraph.
+          If we assess a takedown claim as valid (or are court ordered to take
+          down information), we will check our servers for copies of the
+          relevant content and add it to our block-list for any services related
```
juliangruber, line 332: shouldn't we also say not just to check for copies but also remove them?

juliangruber (changes requested):

juliangruber:
Sweeet

PR #312, opened 2020-10-20
Update: latest sdk resume downloads

dpaez:
[_WIP_] - This PR makes some tweaks to use the latest sdk branch that contains some fixes for resuming downloads and general cloning. It also makes use of the new `addFiles` method.


chartgerink:
I made some final updates, which may make this PR ready to merge. 
1. Upgraded to `@p2pcommons/sdk-js` 0.9.0
2. Take into account older versions of the `p2pcommonsDir` that could be non-writable (gave permission errors)
3. Added a check before launch to ensure that `p2pcommonsDir` exists (e.g., if somebody hard removes the `p2pcommonsDir` like I did, or for some other reason)

PR #335, opened 2020-11-25
Add messageBox for update

### LINKED ISSUE #320: Show changelog before update
**Is your feature request related to a problem? Please describe.**
Updates for Hypergraph are silently downloaded and installed at the moment. This excludes the user from knowing what was updated without active effort.

**Describe the solution you'd like**
Display a dialog that includes the changelog from the release.

**Describe alternatives you've considered**

* Include the changelog in the notification (too long)
* Provide the changelog upon restart

**Additional context**

https://stackoverflow.com/questions/45560775/see-release-notes-before-updating-with-electron-updater

chartgerink:
This changes the update procedure from `autoUpdater.checkForUpdatesAndNotify()` to a dialog box that displays the release name and changelog. For the event and what it returns, see the [`autoUpdater docs`](https://www.electronjs.org/docs/api/auto-updater#event-update-downloaded).﻿ 

This update dialog is more verbose than the system notification, which gives users information about what changes occur in the update. Note: Including a more approachable summary of the update is important for the release notes.

I've tested to the extent that I can, but cannot test this exact implementation for Hypergraph outside of production (due to code signing and release workflow of GitHub) --- I think. Open to suggestions on how to test this if at all possible. I am not 100% on what is returned by `info.releaseName` and `info.releaseNotes` (especially the latter; it's not populated directly but may be grabbed from the GitHub release, otherwise see [`electron-builder` docs](https://www.electron.build/configuration/configuration#metadata)), so it would be helpful to test this prior to merging.

Closes #320.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,7 @@ jobs:
           windows_certs: ${{ secrets.windows_certs }}
           windows_certs_password: ${{ secrets.windows_certs_password }}
           release: ${{ startsWith(github.ref, 'refs/tags/v') }}
+          args: -c.releaseInfo.releaseNotesFile=./release-notes.md
           max_attempts: 5
         env:
           API_KEY_ID: ${{ secrets.api_key_id }}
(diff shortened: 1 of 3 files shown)


chartgerink:
After testing on a fork, it seems like this isn't working at all. This can be revisited whenever it becomes a need. It was educational to try and figure this out nonetheless ☺️

PR #338, opened 2021-01-11
Upgrade electron

chartgerink:
This PR upgrades electron to version 11 (previously v9).

I did get an error when running `npm test` locally, but I wasn't sure whether it was something on my device. Pushing to see how the build performs too.

```
Failed to create session. session not created: This version of ChromeDriver only supports Chrome version 83
```


chartgerink:
Seems like it was an issue on my machine in the end :)

PR #342, opened 2021-01-26
Add sorting to following page

chartgerink:
This PR adds sorting to the `Following` page, so the order is no longer based off of the chronological following (hard to remember) but the alphabetical order of the profile names (easier to find).


app/components/profile/following.js
```diff
@@ -37,13 +37,14 @@ const Following = ({ p2p }) => {
...
-          (await p2p.get(profileUrl)).rawJSON.follows.map(key =>
-            p2p.clone(encode(key))
-          )
+      const x = await Promise.all(
+        (await p2p.get(profileUrl)).rawJSON.follows.map(key =>
+          p2p.clone(encode(key))
         )
       )
+      setFollowing(
+        x.sort((a, b) => (a.rawJSON.title > b.rawJSON.title ? 1 : -1))
```
juliangruber, line 46: `.sort((a, b) => a.localeCompare(b))`

app/components/profile/following.js
```diff
@@ -37,13 +37,14 @@ const Following = ({ p2p }) => {
 
   useEffect(() => {
     ;(async () => {
-      setFollowing(
-        await Promise.all(
-          (await p2p.get(profileUrl)).rawJSON.follows.map(key =>
-            p2p.clone(encode(key))
-          )
+      const x = await Promise.all(
```
juliangruber, line 40: Code is often clearer to read if it doesn't use arbitrary seeming single letter variable names. What about `const following` or `const profiles`?

chartgerink: Thanks. Bad habits from my scripting days, so good call-out 😊

PR #353, opened 2021-02-05
Update window minHeight

### LINKED ISSUE #348: Window can be resized smaller than tour is high
image

chartgerink:
Closes #348.

Now looks like this at the minimum height.

image

@psobrakseaton the old minimum window height was `762px` and I wonder whether there was a reason for that? Pinging you for review so that you can weigh in.


juliangruber:
I had hard coded that value so there would be an even padding when at the minimum window size

juliangruber:
I think we try to stay as small as possible as this might make the app unusable on small displays. See https://screensiz.es/monitor, <800px screen height is common. And even at 800px screen height, there's some OS stuff still, so the actually usable space is smaller.

It's easy to fall into a bias, with developers having new machines and larger screens, but many people use smaller screens.

chartgerink:
> 
> 
> I had hard coded that value so there would be an even padding when at the minimum window size

Thanks for explaining ☺️ 

> 
> 
> I think we try to stay as small as possible as this might make the app unusable on small displays. See https://screensiz.es/monitor, <800px screen height is common. And even at 800px screen height, there's some OS stuff still, so the actually usable space is smaller.
> 
> It's easy to fall into a bias, with developers having new machines and larger screens, but many people use smaller screens.

Very good point, that was a blindspot in my thinking. Thanks for the resource! If increasing the minimum height of the window isn't a good option (which sounds like it), it might be better to shrink the modal minimum height (I tried that first) --- it will require some edits in the modal design/copy.

PR #350, opened 2021-02-05
Add key backup modal

chartgerink:
This PR adds a modal that forces users to back up their key. This is based on the experience that users did not follow on the welcome tour instructions and ended up not creating a backup of their private key. The implemented design is as follows (illustration provided by @psobrakseaton):

Screenshot 2021-02-05 at 10 21 23

Note: This modal does not yet appear for existing users of Hypergraph. There is some commented out code in the pull request that aimed to implement this, but alas, I ran into a wall with how the modals were set up. I ended up paginating to the extent that it hit `if (!dialogs[page]) return null` which logically led to a black screen upon close. Any help appreciated to make existing users also back up their keys :)

PR #357, opened 2021-02-09
Add option to show terms upon update

chartgerink:
This PR adds the option to hardcode reappearance of the terms when updating the application.

Specifically, whenever the terms are updated before release, the `welcome` and `terms` storevalue should be flipped to true. In the PR they are set to `false` given that the terms have not yet been updated in `main`. This will then show the terms to users after the app is updated (plus the final welcome screen). Specifically, it is this tidbit that needs to be changed in `./index.js`: 
```
if (!(app.getVersion() === store.get('lastVersion'))) {
  store.set('welcome', false)
  store.set('terms', false)
}
```

I implemented it by skipping all the pages between the terms and the final welcome screen. This is due to the welcome screens at the moment not being designed for single display (per [this comment](https://github.com/hypergraph-xyz/desktop/pull/350#issuecomment-773917757)). This was my inelegant approach to be able to add it at the moment, but we might want to reconsider the way the welcome screens are setup in the future if this proves too clunky.


app/components/welcome/welcome.js
```diff
@@ -60,107 +60,134 @@ const ButtonGroup = styled.div`
...
-            'analytics',
-            Boolean(data.get('analytics'))
-          )
-          await ipcRenderer.invoke('setStoreValue', 'chatra', true)
-          next()
-        }}
-      >
-        <ButtonGroup>
+  ({ next }) => {
+    return (
```
juliangruber, line 64: unnecessary syntax change

app/components/welcome/welcome.js
```diff
@@ -60,107 +60,134 @@ const ButtonGroup = styled.div`
...
+            await ipcRenderer.invoke(
+              'setStoreValue',
+              'analytics',
+              Boolean(data.get('analytics'))
+            )
+            await ipcRenderer.invoke('setStoreValue', 'chatra', true)
+            await ipcRenderer.invoke('setStoreValue', 'terms', false)
+            await ipcRenderer.invoke(
+              'setStoreValue',
+              'lastVersion',
```
juliangruber, line 80: not clear which version this is, what about `'currentAppVersion'` or `'lastInstalledAppVersion'`?

app/components/welcome/welcome.js
```diff
@@ -60,107 +60,134 @@ const ButtonGroup = styled.div`
...
+          onSubmit={async ev => {
+            ev.preventDefault()
+            const data = new FormData(ev.target)
+            await ipcRenderer.invoke(
+              'setStoreValue',
+              'analytics',
+              Boolean(data.get('analytics'))
+            )
+            await ipcRenderer.invoke('setStoreValue', 'chatra', true)
+            await ipcRenderer.invoke('setStoreValue', 'terms', false)
```
juliangruber, line 77: Why do we set `terms` to `false` here, right after the user saw them? I didn't get that part in the PR description

app/components/welcome/welcome.js
```diff
@@ -60,107 +60,134 @@ const ButtonGroup = styled.div`
...
+            </Button>
+            <Checkbox name='analytics' />
+            Allow analytics to be sent to Liberate Science GmbH
+          </ButtonGroup>
+        </Form>
+      </>
+    )
+  },
+  ({ page, skip, next }) => {
+    if (skip) next()
```
juliangruber, line 99: Instead of each slide manually skipping, the slide manager can simply modify the current page index

juliangruber (changes requested):

app/components/welcome/welcome.js
chartgerink: Good point. `lastInstalledAppVersion` is the best depiction. I updated accordingly :)

app/components/welcome/welcome.js
chartgerink: Hm, the intent is to update it to `false` after the person accepts the terms. `terms` is meant as `showTerms`, which might be the confusing part? I based it off the `welcome` store that functioned more as `showWelcome` too.

app/components/welcome/welcome.js
chartgerink: lol indeed. 👍

app/components/welcome/welcome.js
juliangruber: ah right. Yeah I think that's exactly it!

chartgerink: Updated the store names to be more verbose :)

PR #358, opened 2021-02-14
Add drag out of content files

chartgerink:
This PR adds the dragging out of content files to anywhere on the device for storage. I retrieved the icon from NounProject under my NounPro account and can be used freely. A demo is added below.

Note: Drag out is not supported for other elements at the moment.

https://user-images.githubusercontent.com/2946344/107880054-b8f73000-6edc-11eb-869e-77bbcf66da0a.mov


app/components/content/content.js
```diff
@@ -255,6 +255,14 @@ const Content = ({ p2p, contentKey: key, version, renderRow }) => {
             onClick={() => {
               remote.shell.openPath(`${directory}/${content.rawJSON.main}`)
             }}
+            draggable='true'
```
juliangruber, line 258:
```suggestion
            draggable
```

in HTML, just the attribute being there makes it `true` (can also be changed below)

PR #362, opened 2021-02-19
Refactor colors to @libscie/design-library

chartgerink:
This PR updates the color references according to the newly implemented [`@libscie/design-library`](https://github.com/libscie/design-library). This makes future changes to the color scheme more easily manageable.

The color red was updated in the meantime to be a bit less flashy by @psobrakseaton and that is implemented as a result here too.

PR #363, opened 2021-02-20
Add @libscie/design-library dependency

chartgerink:
This is another attempt to get the checks running. In https://github.com/hypergraph-xyz/desktop/pull/362 the `package-lock.json` was updated to version 2 due to my finally installing `npm@v7` on my machine. Who knows, maybe that is the parameter that caused it to fail (that would still need fixing if it succeeds here).


chartgerink:
Although I still don't understand why there are errors with the cloning of the repositories in the action as a result of `lockFileVersion: 2` 🤷‍♀️

juliangruber:
Did you try updating the npm version for GHA to 7?

chartgerink:
I didn't try that explicitly, good one. Upon trying that it did remove the specific issues I had (thanks!) but the build was stalling/taking a long time. See [here](https://github.com/hypergraph-xyz/desktop/runs/1947240978) --- may be that the build and release needs updating too.

juliangruber:
Hm yeah could be, but probably it's safer to just wait a bit and let the ecosystem fully implement npm 7

PR #364, opened 2021-02-21
Set default content type

### LINKED ISSUE #361: Edit content resets content type
## Description

When editing a draft research module, the content type resets to `Idea`. 

## To Reproduce

1. Add draft module with content type different from `Idea`
2. Click `Edit content`
3. Check content type

## Expected behavior

Passes the already selected option to the edit content screen.

## Actual behavior

Passes the default option for creating a new research module.

## System

- OS (incl. version): macOS 11.2.1
- Hypergraph Desktop version: v0.9.0

chartgerink:
This PR adds a check for an existing content type when editing a research module. In case there is a content type passed in to the form, it selects that by default. If no content type is available already (e.g., when creating a new research module), it defaults to [the first item in the list](https://github.com/hypergraph-xyz/wikidata-identifiers/blob/main/index.json).

Closes #361.


app/components/content/edit-form.js
```diff
@@ -384,7 +384,13 @@ const EditForm = ({
         <NewSelect
           className='basic-single'
           classNamePrefix='select'
-          defaultValue={options[0]}
+          defaultValue={
+            !subtype
+              ? options[0]
+              : options.filter(x => {
+                if (x.value === subtype) return x
+              })[0]
```
juliangruber, line 392:
```suggestion
              : options.find(x => x.value === subtype)
```

Simpler using `Array#find`

chartgerink: Thank you for sharing your knowledge 😊 (helps me learn these methods too)

PR #366, opened 2021-02-22
Update key backup notifications

chartgerink:
The key backup notification showed upon first launch, which was redundant because of the welcome screens now forcing key backup. The notification may be potentially a bit annoying too (I found it annoying). 

This PR removes the notification when the application is launched for the first time.


index.js
```diff
@@ -67,7 +67,12 @@ ipcMain.handle('dragOut', (_, value) => {
   })
 })
 
-if (!(app.getVersion() === store.get('lastInstalledAppVersion'))) {
+if (
+  !(app.getVersion() === store.get('lastInstalledAppVersion')) &&
```
juliangruber, line 71:
```suggestion
  app.getVersion() !== store.get('lastInstalledAppVersion') &&
```

Easier to reason about without the double logic

PR #372, opened 2021-02-24
Add security check for file opening

chartgerink:
Files without extension would be opened in the shell (at least on macOS), which is an attack vector for potential exploits. Unknowing users might try to open up a file, and it could actually be a (malicious) script that does something completely different.

This PR adds a security check and alerts users if there is no extension. It gives them the option to open the folder and inspect manually. We may want to add a check at the other end in the content form, but that's a UX item and not a security fix (pinging @psobrakseaton).


app/components/content/content.js
```diff
@@ -140,6 +140,24 @@ const getContentDirectory = async ({ p2p, key, version }) => {
   return version ? `${directory}+${version}` : directory
 }
 
+const secureFileOpen = (dir, file) => {
+  const extensionPresent = new RegExp(/(\..*)$/)
```
juliangruber, line 144: safer to use https://nodejs.org/api/path.html#path_path_extname_path

app/components/content/content.js
```diff
@@ -140,6 +140,24 @@ const getContentDirectory = async ({ p2p, key, version }) => {
...
+  if (!extensionPresent.test(file)) {
+    const choice = remote.dialog.showMessageBoxSync({
+      type: 'question',
+      title: 'Security',
+      message:
+        'For your security, Hypergraph does not open files without extensions. Please inspect file manually.',
```
juliangruber, line 150: What if it does have an extension, and that is something like `.sh`, which would also be a script?

juliangruber (changes requested):

app/components/content/content.js
chartgerink: Good point! I tested this on my macOS device and didn't run into that issue (`.sh` opened in editor) --- hence, I decided to leave it for now as it seems like this works (more like) expected. The files without extensions ended up opening up the shell unexpectedly.

juliangruber: Have you tried on a mac that doesn't have a developer text editor? Or what about `.bat` files or `.exe` files on Windows?

chartgerink: Ah yes of course 😅 `.exe` is not permitted to be the main file, so that shouldn't even come up if all the validations work properly in the @p2pcommons/sdk-js. I'll add a few conditions to cover `.sh|.bat|.ps1`

PR #384, opened 2021-02-28
Update index.js

chartgerink:
Fixes the protocol for the redirect domain http://hypergraph.xyz (old: https://hypergraph.xyz).


juliangruber:
What's the issue with that domain not working with https?

chartgerink:
Redirect, and it didn't allow for https through the domain registrar.

PR #395, opened 2021-03-01
Add isFile condition

### LINKED ISSUE #390: Security warning displayed when opening folder
I tried to open the `data` folder:

image

Can use `fs.stat()` to check if it's a directory

chartgerink:
This PR adds a check that `secureFileOpen` only triggers when a file is being openend (not triggered for a folder).

Closes #390.


app/components/content/content.js
```diff
@@ -142,10 +142,11 @@ const getContentDirectory = async ({ p2p, key, version }) => {
   return version ? `${directory}+${version}` : directory
 }
 
-const secureFileOpen = (dir, file) => {
+const secureFileOpen = async (dir, file) => {
+  const isFile = await (await fs.stat(`${dir}/${file}`)).isFile()
```
juliangruber, line 146:
```suggestion
  const isFile = (await fs.stat(`${dir}/${file}`)).isFile()
```

juliangruber (changes requested):

app/components/content/content.js
chartgerink: The linter keeps adding them when I do it 😅 Thanks for pointing it out.

juliangruber: Interesting, how exactly is your code before the formatter runs?

chartgerink: Oh wait I think it was my code completion in VSCode, not the linter. It for some reason adds it when I add the .isFile method (maybe because the `fs` is imported as promise?).

Issue #297, opened 2020-09-28
Main file not auto-removed from index.json upon deletion


chartgerink:
## Description

When a file is removed from a draft, and that was the designated main file, the `index.json` isn't updated to reflect there is no longer a main file. When saving this draft, it also hangs because the promise of an existing main file doesn't get fulfilled.

## To Reproduce

1. Add a draft with a main file
2. Click Edit content
3. Remove the file with the X
4. Save as draft

## Expected behavior

The `index.json` gets updated to reflect the main file was removed.

## Actual behavior

The `index.json` does not get updated and refers to a no longer existing file.

## Screenshots

![Screen Recording 2020-09-28 at 14 27 47 2020-09-28 14_36_51](https://user-images.githubusercontent.com/2946344/94433141-23ab0780-0198-11eb-9c1a-f5097a785d07.gif)

## System

- OS (incl. version): macOS 10.15.6
- Hypergraph Desktop version: 2cc9ac459d469c0c83788b9f37c44e8e0baf304a

juliangruber:
We can't yet update the `index.json` automatically, as we don't have events for file changes in the sdk, however I can unbreak the edit content screen.

PR #405, opened 2021-03-20
Add file drag-and-drop

chartgerink:
This PR adds a drag-and-drop for files in the add (and edit) content screen. The basic implementation is demonstrated in the videos below. Lint keeps at it for the README, too. 🤷‍♀️ 

### Reactive styling

https://user-images.githubusercontent.com/2946344/111883250-68cb2c00-89ba-11eb-90a4-d229b5f83d8e.mov

### Dialog box

https://user-images.githubusercontent.com/2946344/111883429-85b42f00-89bb-11eb-8401-3ad3ef96f95f.mov

### Limitations/context

* Changing the dropstyle seems to be  based on the MIME type only. This is why there is no reject styling, as it would accept files but style them as reject overly liberally.
* #297 still occurs (as expected but I forgot at first while testing 😅 )

### Next steps

- [x] Designs dropbox
  - [x] General state (purple now)
  - [x] Accept state (green now)
  - [x] Ambiguous state (yellow now)
- [x] Text for dropbox
- [x] Additional testing
  - [x] macOS
  - [x] Linux
  - [x] Windows


psobrakseaton:
So this is pretty much perfect except I feel that the height of the box should be changed since with this change we will have 4 different heights in the form and it feels messy. Could we make it so that this is the same size as the 'Description' box?

psobrakseaton:
> So this is pretty much perfect except I feel that the height of the box should be changed since with this change we will have 4 different heights in the form and it feels messy. Could we make it so that this is the same size as the 'Description' box?

This appears to be 188px

chartgerink:
Thanks @psobrakseaton! Here is a screenshot of the implementation after coordinating on mattermost about some of the margins.

image

I just need to test for a bit on linux and macOS on my end.

PR #414, opened 2021-03-24
Change release asset names

chartgerink:
This PR removes the version number from the release assets, and adds an about button to the `Help` menu. 

This allows easier download links across website(s) with links like this.

`https://github.com/hypergraph-xyz/desktop/releases/latest/download/hypergraph-desktop.dmg`


juliangruber:
Have you tried if auto updates still work like this?

If yes, then that's a great improvement

index.js
```diff
@@ -21,6 +21,7 @@ const Store = require('electron-store')
 const { autoUpdater } = require('electron-updater')
 const log = require('electron-log')
 const express = require('express')
+const { version: p2pcommons } = require('@p2pcommons/sdk-js/package.json')
```
juliangruber, line 24: This should work, but it would be cleaner if `p2pcommons` exposed the version itself, like

```js
const { version: p2pcommons } = require('@p2pcommons/sdk-js')
```

index.js
```diff
@@ -201,6 +202,15 @@ const updateMenu = () => {
...
         submenu: [
+          {
+            label: 'About',
+            click: () => {
+              dialog.showMessageBox({
+                message: `Running version ${app.getVersion()} of ${app.getName()}\nPowered by version ${p2pcommons} of the p2pcommons`,
+                type: 'info'
+              })
+            }
+          },
```
juliangruber, line 213: This is very useful 👍

chartgerink:
Ah yes, thank you for mentioning the testing. 

I tried it in [a fork](https://github.com/chartgerink/desktop) and it seems to work as expected. :) That is (for the fork): if you install [`v0.10.1`](https://github.com/chartgerink/desktop/releases/tag/v0.10.1) it should update to [`v0.10.2`](https://github.com/chartgerink/desktop/releases/tag/v0.10.2) with the new asset names).

PR #415, opened 2021-03-24
Revert to single click

### LINKED ISSUE #386: Double clicking on a module row
First, I found it unexpected that you needed to double click on a content row to open it, but then I remembered this was changed.

I did notice however that when you double click a content row it for a second selects the content type, before navigating.

image

chartgerink:
This PR reverts to content rows being single click to open. Also deactivates the text selection on the content cards. This was the desired end result as indicated by @psobrakseaton for the UX.

Closes #386.


juliangruber (approved):
✨

